PR #10, opened 2022-02-22
Ien 36 spreadsheet upload

coryatwater-FW:
I added just a upload option for spreadsheets, it just logs the data to the console so we can see what it looks like and decide how we want to handle it


dbayly-freshworks:
We should rename `login.tsx` to something else since we aren't making a login page anymore.

PR #16, opened 2022-02-24
Ien 36 store form in database

dbayly-freshworks:
Added a form module to allow the storage of the json data of form. 

Also fixed a lot of the pipeline issues that we've been facing. 

I am thinking of removing the post to slack integration as it's currently the source of a lot of noise.

### DIFF
--- a/apps/api/src/form/form.controller.ts
+++ b/apps/api/src/form/form.controller.ts
@@ -1,4 +1,4 @@
-import { SubmissionDTO } from '@ien/common';
+import { FormDTO } from '@ien/common/src/dto';
 import {
   Inject,
   Logger,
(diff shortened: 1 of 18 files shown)


bhumin-fw (approved):

apps/api/src/form/form.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+  ) {}
+
+  @ApiOperation({
+    summary: 'Insert a new Form',
+  })
+  @UseInterceptors(ClassSerializerInterceptor)
+  @ApiResponse({ status: HttpStatus.CREATED, type: EmptyResponse })
+  @HttpCode(HttpStatus.CREATED)
+  @Post()
+  async name(@Body() body: FormDTO): Promise<any> {
```
bhumin-fw, line 33: should replace **any** => **FormEntity** and return created object?

coryatwater-FW (approved):

PR #18, opened 2022-02-28
IEN 58 add applicant

bhumin-fw:
Add applicant API
Update Applicant and status API

### DIFF
--- a/apps/api/src/applicant/applicant.controller.ts
+++ b/apps/api/src/applicant/applicant.controller.ts
@@ -55,6 +55,7 @@ export class ApplicantController {
   @ApiOperation({
     summary: 'Get Applicant details',
   })
+  @ApiQuery({ name: 'relation', required: false, description: 'Get additinal data like audit' })
   @UseInterceptors(ClassSerializerInterceptor)
   @ApiResponse({ status: HttpStatus.OK, type: EmptyResponse })
   @HttpCode(HttpStatus.OK)
--- a/apps/api/src/applicant/applicant.controller.ts
+++ b/apps/api/src/applicant/applicant.controller.ts
@@ -19,10 +19,10 @@ import { ApplicantService } from './applicant.service';
 import { EmptyResponse } from 'src/common/ro/empty-response.ro';
 import { ApplicantEntity } from './entity/applicant.entity';
 import { AppLogger } from 'src/common/logger.service';
-import { ApplicantCreateRO } from './ro/applicant-create.ro';
-import { ApplicantFilterByIdRO } from './ro/applicant-by-id.ro';
-import { ApplicantFilterRO } from './ro/applicant-filter.ro';
-import { ApplicantUpdateRO } from './ro/applicant-update.ro';
+import { ApplicantCreateAPIDTO } from './dto/applicant-create.dto';
+import { ApplicantFilterByIdAPIDTO } from './dto/applicant-by-id.dto';
+import { ApplicantFilterAPIDTO } from './dto/applicant-filter.dto';
+import { ApplicantUpdateAPIDTO } from './dto/applicant-update.dto';
 
 @Controller('applicant')
 @ApiTags('Applicant')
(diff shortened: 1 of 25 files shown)


apps/api/src/applicant/applicant.controller.ts
```diff
@@ -47,11 +55,44 @@ export class ApplicantController {
   @ApiOperation({
     summary: 'Get Applicant details',
   })
+  @ApiQuery({ name: 'relation', required: false, description: 'Get additinal data like audit' })
```
coryatwater-FW, line 58: small typo on additional*

apps/api/src/applicant/applicant.service.ts
```diff
@@ -50,4 +72,166 @@ export class ApplicantService {
...
+    return statusObj;
+  }
+
+  /**
+   * Add new applicant data in database table applicant data
+   * @param data Applicant DTO
+   * @param statusObj Applicant status Object
+   * @returns Applicant or Not Found
+   */
+  async saveAapplicant(
```
coryatwater-FW, line 139: small typo here

packages/common/src/dto/applicant-create.dto.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  @IsString()
+  @IsNotEmpty()
+  @Length(1, 256)
+  profession!: string;
+
+  @IsString()
+  @Length(1, 256)
+  @IsOptional()
+  speciality?: string;
```
coryatwater-FW, line 31: I've noticed two different naming conventions for speciality (spreadsheet uses 'specialty'), I assume we should be using 'specialty' but we should decide 100% which we'll be moving forward with,

apps/api/src/applicant/applicant.controller.ts
```diff
@@ -8,11 +9,18 @@ import {
...
   Query,
   UseInterceptors,
 } from '@nestjs/common';
 import { ApiOperation, ApiResponse, ApiTags, ApiQuery } from '@nestjs/swagger';
-import { ApplicantFilterDto } from '@ien/common';
+import {
+  ApplicantFilterDto,
+  ApplicantCreateDto,
+  ApplicantUpdateDto,
+  ApplicantFilterByIdDto,
```
coryatwater-FW, line 22: noticed two different naming conventions for DTO's (exampleDTO and exampleDto).  we should decide on one and stick with it

packages/common/src/dto/applicant-create.dto.ts
bhumin-fw: Agree we will go with ‘specialty’ only. Let me update it.

apps/api/src/applicant/applicant.controller.ts
bhumin-fw: Updating it to, exampleDTO

apps/api/src/applicant/applicant.controller.ts
```diff
@@ -29,15 +35,13 @@ export class ApplicantController {
...
     summary: 'List applicants',
   })
-  @ApiQuery({ name: 'ha_pcn', required: false, description: 'Provide optional HA' })
-  @ApiQuery({ name: 'status', required: false, description: 'Provide optional status(int)' })
   @UseInterceptors(ClassSerializerInterceptor)
   @ApiResponse({ status: HttpStatus.OK, type: EmptyResponse })
   @HttpCode(HttpStatus.OK)
   @Get()
-  async getApplicants(@Query() filterDto: ApplicantFilterDto): Promise<ApplicantEntity[]> {
+  async getApplicants(@Query() filter: ApplicantFilterRO): Promise<ApplicantEntity[]> {
```
dbayly-freshworks, line 42: Shouldn't this accept an ApplicantFilterDTO and return an array of ApplicantRO?

apps/api/src/applicant/applicant.controller.ts
```diff
@@ -51,7 +55,39 @@ export class ApplicantController {
...
+  }
+
+  @ApiOperation({
+    summary: 'Add Applicant',
+  })
+  @UseInterceptors(ClassSerializerInterceptor)
+  @ApiResponse({ status: HttpStatus.CREATED, type: EmptyResponse })
+  @HttpCode(HttpStatus.CREATED)
+  @Post()
+  async addApplicant(@Body() addApplicant: ApplicantCreateRO): Promise<ApplicantEntity> {
```
dbayly-freshworks, line 72: Nit - We should return and RO instead of an entity so we can limit the fields the client receives. 
Usually the pattern is to fetch the Entity and then call a `toResponseObject` function on the entity.

dbayly-freshworks (commented):
A few small comments but overall this looks good.

dbayly-freshworks (approved):

PR #22, opened 2022-03-04
IEN-5 add new IEN

coryatwater-FW:


apps/web/src/services/constants/applicant-details-constants.ts
```diff
@@ -0,0 +1,95 @@
...
+  { value: 'midwife', label: 'Midwife' },
+  { value: 'nurse', label: 'Nurse' },
+  { value: 'occupational-therapist', label: 'Occupational Therapist' },
+  { value: 'physiotherapist', label: 'Physiotherapist' },
+  { value: 'pharmacist', label: 'Pharmacist' },
+  { value: 'sonographer', label: 'Sonographer' },
+];
+
+export const specialityOptionsArray = [
+  { value: 'chronic-disease', label: 'Chronic Disease' },
```
dbayly-freshworks, line 16: I'm wondering about the pattern here. Since we have so many, it might make sense to have lower-kebab-case for the values.

dbayly-freshworks: Also keeping this consistent over all the options as well is important.

dbayly-freshworks (commented):
Looks good. Just a few questions around values and naming conventions.

PR #30, opened 2022-03-21
add new record

coryatwater-FW:


apps/web/src/components/display/AddRecordModal.tsx
```diff
@@ -1,23 +1,38 @@
...
-export const AddRecordModal: React.FC = () => {
+interface AddRecordProps {
+  jobRecords: any;
+  setJobRecords: any;
+}
+
+export const AddRecordModal: React.FC<AddRecordProps> = ({ jobRecords, setJobRecords }) => {
   const router = useRouter();
 
+  const applicantId = router.query.applicantId;
```
dbayly-freshworks, line 19: Just to be safe, it might be good to add optional chaining  here just in case a user ends up on this page without an id.

dbayly-freshworks (approved):
This looks good for now. I agree that caching later will be useful to prevent unnecessary useEffect triggers.

PR #33, opened 2022-03-25
Remove all old applicant reference that are not in-use.

bhumin-fw:


dbayly-freshworks (approved):
Happy to see this all cleaned out

PR #37, opened 2022-03-30
IEN-110 Import master table data from HMBC ATS to IEN app

bhumin-fw:
- Add a module that fetches data from HMBC ATS URL, and saves it in our database.
- In this PR only consider Open API that does not require authentication.
- Change Primary Key auto generate to Primary key only for better upsert function with  ORM.

### DIFF
--- a/apps/api/src/applicant/entity/ienstatus-reason.entity.ts
+++ b/apps/api/src/applicant/entity/ienstatus-reason.entity.ts
@@ -1,0 +1,10 @@
+import { Entity, Column, PrimaryColumn } from 'typeorm';
+
+@Entity('ien_status_reasons')
+export class IENStatusReason {
+  @PrimaryColumn()
+  id!: number;
+
+  @Column('varchar')
+  name!: string;
+}
--- a/apps/api/src/applicant/external-api.service.ts
+++ b/apps/api/src/applicant/external-api.service.ts
@@ -1,16 +1,18 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { Inject, Injectable } from '@nestjs/common';
+import { Inject, Injectable, Logger } from '@nestjs/common';
 import { ExternalRequest } from 'src/common/external-request';
 import { InjectRepository } from '@nestjs/typeorm';
 import { IENHaPcn } from './entity/ienhapcn.entity';
 import { Repository } from 'typeorm';
 import { IENUsers } from './entity/ienusers.entity';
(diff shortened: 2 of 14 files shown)


apps/api/src/applicant/external-api.service.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Collect Milestone reason data from <domain>/Reason Url.
+   * and Save in ien_status_reasons table in database for milestone/status reference.
+   */
+  async saveReasons(): Promise<void> {
+    const data = await this.external_request.getReason();
+    if (data instanceof Array) {
+      const keys: any = this.cleanData(data);
+      console.log({ keys });
```
dbayly-freshworks, line 93: Stray console.log

apps/api/src/applicant/entity/ienstatus-reason.entity.ts
```diff
@@ -0,0 +1,10 @@
+import { Entity, Column, PrimaryColumn } from 'typeorm';
+
+@Entity('ien_status_reasons')
+export class IENStatusReason {
+  @PrimaryColumn()
+  id!: number;
+
+  @Column('varchar')
+  name!: string;
```
dbayly-freshworks, line 9: Does this need to be a separate table? Can this exist as an enum?

apps/api/src/applicant/external-api.service.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Save all data for master tables.
+   */
+  async saveData() {
+    const ha = this.saveHa();
+    const users = this.saveUsers();
+    const reasons = this.saveReasons();
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    await Promise.all([ha, users, reasons]).then(res => console.log('done'));
```
dbayly-freshworks, line 31: Maybe use logger instead of console.log here?

apps/api/src/common/external-request.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Injectable()
+export class ExternalRequest {
+  private baseurl: string;
+  private api_instance;
+  constructor() {
+    this.baseurl = process.env.HMBC_ATS_BASE_URL || `https://ien.heabc.bc.ca`;
+    this.api_instance = axios.create({
+      baseURL: this.baseurl,
+      timeout: 15000,
```
dbayly-freshworks, line 13: 15 seconds seems very generous. What scale of data are we expecting?

PR #44, opened 2022-04-08
IEN-211 seed data and response object for all api endpoints

bhumin-fw:


apps/api/src/applicant/external-api.service.ts
```diff
@@ -180,13 +180,20 @@ export class ExternalAPIService {
           });
         }
 
+        let citizenship = null;
+        if (a.country_of_citizenship && a.country_of_citizenship instanceof Array) {
```
sunghwan-park-fw, line 184: How about `Array.isArray(a.country_of_citizenship)`?

sunghwan-park-fw (approved):

PR #48, opened 2022-04-11
add husky precommit

coryatwater-FW:


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn precommit
```
sunghwan-park-fw, line 4: if we don't use precommit script, how about deleting it from package.json?

dbayly-freshworks (approved):
Tried this out locally and it works great!

.husky/pre-commit
coryatwater-FW: good catch! done

PR #50, opened 2022-04-12
Fixed app deployment

dbayly-freshworks:
While testing out my work on keycloak I noticed my changes to the web app were not being reflected.
It looks like `make deploy app` was just invalidating the cloudfront distribution. `sync-app` should actually deploy the front end.


bhumin-fw (approved):
Go ahead

PR #51, opened 2022-04-12
added RO types to replace any types, add try catch and toast in service

coryatwater-FW:


apps/web/src/components/milestone-logs/recruitment/Record.tsx
```diff
@@ -90,7 +81,7 @@ export const Record: React.FC<RecordProps> = ({ job }) => {
               <DetailsItem title='Recruiter Name' text={recruiter_name} />
               <DetailsItem
                 title='Date Job Was First Posted'
-                text={formatDate(job_post_date.toString())}
+                text={job_post_date && formatDate(job_post_date)}
```
sunghwan-park-fw, line 84: formatDate might handle null and undefined.

coryatwater-FW: you are correct, I will remove those checks

apps/web/src/components/milestone-logs/recruitment/Milestone.tsx
```diff
@@ -46,14 +45,10 @@ export const AddMilestones: React.FC<AddMilestoneProps> = ({
     values.job_id = jobId.toString();
     values.added_by = '1';
 
-    try {
-      const {
-        data: { data },
-      } = await addMilestone(applicantId as string, values);
+    const data = await addMilestone(applicantId as string, values);
```
sunghwan-park-fw, line 48: applicantId's type is `string | string[] | undefined`. should we change it to `string` or check if it is string here?

coryatwater-FW: changed it to string in parent comp so we can change type to just string

packages/common/src/helper/jsonToArray.ts
```diff
@@ -0,0 +1,11 @@
+// used for converting json values to an array to allow map fn to work
+export const jsonToArray = (json?: any) => {
```
sunghwan-park-fw, line 2: isn't it same with Object.values(json)?

sunghwan-park-fw (approved):
I left some comments.

packages/common/src/helper/jsonToArray.ts
coryatwater-FW: yes, I changed it to Object.values since I have no need for the keys, good catch!

PR #52, opened 2022-04-12
fix pending details page

sunghwan-park-fw:


apps/web/src/pages/details/[applicantId].tsx
```diff
@@ -82,11 +82,7 @@ const Details = () => {
         <div className='col-span-3'>
           <DetailsItem
             title='Assigned To'
-            text={
-              applicant.assigned_to
-                ? applicant.assigned_to.map((a: { name: any }) => a.name).join(',')
-                : 'NA'
-            }
+            text={applicant.assigned_to?.map((a: { name: any }) => a.name).join(',') || 'NA'}
```
Comment (bhumin-fw) [inline apps/web/src/pages/details/[applicantId].tsx:85]: can we replace 'any' with 'string' here(' name: any')?

bhumin-fw (approved):
One more point if you can add in this PR.
country_of_citizenship is a **string[]** and not a **string** anymore. so now you will receive value like ['ca'] or ['it', 'ca']. just need to apply join() while displaying these values.

Please go ahead and merge once you done with these points.

apps/web/src/pages/details/[applicantId].tsx
Comment (coryatwater-FW) [reply apps/web/src/pages/details/[applicantId].tsx:85]: both those were fixed in my most recent PR

Comment (sunghwan-park-fw) [reply apps/web/src/pages/details/[applicantId].tsx]: @bhumin-fw 
`nursing_educations`, `health_authorities`, and `assigned_to` seem to be an array, so how about changing their type 
from JSON to array of their type?

Comment (bhumin-fw) [reply apps/web/src/pages/details/[applicantId].tsx]: @sunghwan-park-fw That's a good idea. We can put an array of objects. You can update it in this PR, or I can push the same in my next PR whatever is suitable for you.

Comment (sunghwan-park-fw) [reply apps/web/src/pages/details/[applicantId].tsx]: Please, update RO then I will update components referencing the fields.

PR #55, opened 2022-04-13
Update Applicant status, Name wise search enhance

bhumin-fw:
- Update Applicant status.
- Name-wise search enhancement.

### DIFF
--- a/apps/api/src/applicant/ienapplicant.service.ts
+++ b/apps/api/src/applicant/ienapplicant.service.ts
@@ -278,7 +278,9 @@ export class IENApplicantService {
     status_id: string,
     applicantUpdate: IENApplicantUpdateStatusAPIDTO,
   ): Promise<IENApplicantStatusAudit | any> {
-    const status_audit = await this.ienapplicantStatusAuditRepository.findOne(status_id, {relations: ['applicant']});
+    const status_audit = await this.ienapplicantStatusAuditRepository.findOne(status_id, {
+      relations: ['applicant'],
+    });
     if (!status_audit) {
       throw new NotFoundException('Provided status/milestone record not found');
     }
@@ -307,9 +309,9 @@ export class IENApplicantService {
       status_audit.notes = notes;
     }
     await this.ienapplicantStatusAuditRepository.save(status_audit);
-    
-     // Let's check and updated latest status on applicant
-     await this.ienapplicantUtilService.updateLatestStatusOnApplicant([status_audit.applicant.id]);
+
+    // Let's check and updated latest status on applicant
+    await this.ienapplicantUtilService.updateLatestStatusOnApplicant([status_audit.applicant.id]);
 
     return status_audit;
   }
--- a/apps/api/src/applicant/ienapplicant.service.ts
+++ b/apps/api/src/applicant/ienapplicant.service.ts
@@ -261,6 +261,9 @@ export class IENApplicantService {
       await this.ienapplicantUtilService.updatePreviousActiveStatusForJob(job, data);
     }
 
+    // Let's check and updated latest status on applicant
+    await this.ienapplicantUtilService.updateLatestStatusOnApplicant([applicant.id]);
+
     delete status_audit.applicant;
     return status_audit;
   }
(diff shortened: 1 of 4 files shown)


apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -63,43 +95,80 @@ export class IENApplicantUtilService {
...
           return `health_authorities @> '[{"id":${id}}]'`;
         })
-        .join(' or ');
-      ha_search_sql = `(${ha_search_sql})`;
+        .join(' OR ');
+      whereHa = `(${whereHa})`;
+      isHaPcn = true;
+    }
 
+    if (isName && isStatus && isHaPcn) {
```
sunghwan-park-fw, line 126: I tried this.
```typescript

    const conditions: (string|ObjectLiteral)[] = [];

    if (status) {
      const status_list = await this.fetchChildStatusList(status);
      conditions.push({ status: In(status_list) });
    }

    if (name) {
      conditions.push(this._nameSearchQuery(name));
    }

    if (ha_pcn) {
      const ha_pcn_array = ha_pcn?.split(',');
      const condition = ha_pcn_array
        ?.map(id => {
          return `health_authorities @> '[{"id":${id}}]'`;
        })
        .join(' OR ');
      conditions.push(`(${condition})`);
    }

    return this.ienapplicantRepository.findAndCount({
      where: (qb: SelectQueryBuilder<IENApplicant>) => {
        const condition = conditions.shift();
        if (condition) qb.where(condition);
        conditions.forEach(condition => qb.andWhere(condition));
      },
      ...query,
    });
```

apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -317,4 +386,17 @@ export class IENApplicantUtilService {
     }
     return job_title;
   }
+
+  async updateLatestStatusOnApplicant(mappedApplicantList: string[]): Promise<void> {
+    try {
+      // update applicant with latest status
+      const idsToUpdate = `'${mappedApplicantList.join("','")}'`;
```
sunghwan-park-fw, line 393: how about using a view for query so that we can remove status field from applicant table? it might make updating queries simpler.

sunghwan-park-fw (commented):
I left some suggestions.

apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -317,4 +386,17 @@ export class IENApplicantUtilService {
...
+      const queryToUpdate = `UPDATE ien_applicants SET status_id = (SELECT status_id FROM ien_applicant_status_audit asa WHERE asa.applicant_id=ien_applicants.id ORDER BY asa.start_date DESC limit 1) WHERE ien_applicants.id IN (${idsToUpdate})`;
```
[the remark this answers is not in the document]
bhumin-fw: If you see a query, I have added a subquery with order and limit value. I was not able to do it in querybuilder for FK relation field.
As I need to fetch status with latest date, it could be added by us or sync from and HMBC ATS.

I have tried below code... and some other possible combination that we can apply. But no success it update null value.

.createQueryBuilder()
        .update(IENApplicant)
        .set({
          status: () =>
            '(SELECT status_id FROM ien_applicant_status_audit sa WHERE sa.applicant_id=ien_applicants.id ORDER BY asa.start_date DESC limit 1)',
        })
        .where('id IN (:...id)', { id: mappedApplicantList })
        .execute();

apps/api/src/applicant/ienapplicant.util.service.ts
bhumin-fw: I was above pushing one simple way, but your approach looks cleaner. Much appreciated the suggestion here. As I was about to ask you for the same in our group.

apps/api/src/applicant/ienapplicant.service.ts
```diff
@@ -304,6 +309,10 @@ export class IENApplicantService {
       status_audit.notes = notes;
     }
     await this.ienapplicantStatusAuditRepository.save(status_audit);
+
+    // Let's check and updated latest status on applicant
+    await this.ienapplicantUtilService.updateLatestStatusOnApplicant([status_audit.applicant.id]);
```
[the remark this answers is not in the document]
bhumin-fw: We do not need it here, we are prosessiong only valid data and due to that update query won't fail. As we are not passing any user specified value in it.

sunghwan-park-fw (approved):

PR #59, opened 2022-04-14
IEN-217 | add milestone table

sunghwan-park-fw:


apps/web/src/components/milestone-logs/MilestoneTable.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <tbody className='text-bcBlack'>
+            {audits.map(audit => (
+              <tr
+                key={audit.id}
+                className='text-left text-sm even:bg-bcLightGray shadow-xs whitespace-nowrap'
+              >
+                <td className='pl-8 py-5 max-w-xs'>{getMilestone(audit)}</td>
+                <td className='px-4'>{formatDate(audit.start_date || '')}</td>
+                <td className='px-4'>{formatDate(audit.end_date || '')}</td>
+                <td className='px-4'>{audit.status_period} days</td>
```
bhumin-fw, line 64: Let's set status-period '' (blank) for now (or if the end-date is not null show status period else blank). As these are data coming from the HMBC system and they do not have an end date. Later once we do receive production data will confirm again and check it again and if still the same, then we will write custom logic on our end to calculate the period of any individual status.

bhumin-fw (approved):
Added one comment for status-period. Please update it other than that it looks good.

PR #61, opened 2022-04-18
IEN 209 implement roles guard

dbayly-freshworks:
Added a roles guard
To create a new route, add both the UseGuards and the RouteAcceptsRoles decorators to your route. 

Also updated the response lambda function to accept responses from the new keycloak origin

### DIFF
--- a/apps/api/src/applicant/ienapplicant.controller.ts
+++ b/apps/api/src/applicant/ienapplicant.controller.ts
@@ -36,6 +36,7 @@ import { RouteAcceptsRoles } from 'src/common/decorators';
 
 @Controller('ien')
 @ApiTags('IEN Applicant')
+@UseGuards(AuthGuard)
 export class IENApplicantController {
   constructor(
     @Inject(Logger) private readonly logger: AppLogger,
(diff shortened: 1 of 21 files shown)


apps/api/src/applicant/ienapplicant.controller.ts
```diff
@@ -43,6 +47,12 @@ export class IENApplicantController {
   })
   @UseInterceptors(ClassSerializerInterceptor)
   @ApiResponse({ status: HttpStatus.OK, type: EmptyResponse })
+  @RouteAcceptsRoles(
+    ValidRoles.HEALTH_AUTHORITY,
+    ValidRoles.HEALTH_MATCH,
+    ValidRoles.MINISTRY_OF_HEALTH,
+  )
+  @UseGuards(AuthGuard)
```
bhumin-fw, line 55: can we move @UserGuards to the controller level?
like this
@Controller('ien')
@UseGuards(AuthGuard)

and last 2 function in the same file missing @RouteAcceptRoles.

bhumin-fw (approved):
left one comment. Other than that it looks good to me.

apps/api/src/applicant/ienapplicant.controller.ts
dbayly-freshworks: This is a good idea, I will make the changes

PR #65, opened 2022-04-20
Add lambda and scheduler to run sync API everyday

bhumin-fw:


Makefile
```diff
@@ -222,6 +223,13 @@ build-web:
 	@mv ./apps/web/out ./terraform/build/app
 	@echo "++\n*****"
 
+build-lambda:
+	@echo "++\n***** Build Lambda for AWS\n++"
+	@mkdir -p ./terraform/lambda
+	@echo 'Copy lambda ...\n' && cp -r ./apps/lambda/* ./terraform/lambda/
+	@echo 'Creating Zip ...\n' && cd ./terraform/lambda/import-applicants && zip -r import-applicants.zip ./* && cd ../..
```
vyasworks, line 230: Not sure if we need to zip the file.

Makefile
```diff
@@ -264,7 +272,10 @@ deploy-app:
 deploy-api:
 	aws lambda update-function-code --function-name ien-$(ENV_NAME)-api --zip-file fileb://./terraform/build/api.zip --region $(AWS_REGION)
 
-deploy-all: sync-app deploy-api
+deploy-lambda:
+	aws lambda update-function-code --function-name sync-applicants --zip-file fileb://./terraform/lambda/import-applicants/import-applicants.zip --region $(AWS_REGION)
```
vyasworks, line 276: Function name should have the ENV at the end

vyasworks: Build artifacts should be only in `/build`

terraform/lambdafunction.tf
```diff
@@ -0,0 +1,63 @@
+resource "aws_lambda_function" "SyncApplicants" {
+  description                    = "Trigger Sync applicant data API"
+  function_name                  = "sync-applicants"
```
vyasworks, line 3: use same patter as other lambda for name

terraform/lambdafunction.tf
```diff
@@ -0,0 +1,63 @@
...
+  role                           = aws_iam_role.lambda.arn
+  runtime                        = "nodejs14.x"
+  filename                       = var.sync_applicants_js
+  source_code_hash               = filebase64sha256(var.sync_applicants_js)
+  handler                        = "index.handler"
+  memory_size                    = 2048
```
vyasworks, line 9: Does this require  2GB memory?

terraform/lambdafunction.tf
```diff
@@ -0,0 +1,63 @@
...
+  timeout                        = 450
```
vyasworks, line 10: Is this 450 ms or seconds. 

It should be less than 30 seconds ideally

terraform/lambdafunction.tf
```diff
@@ -0,0 +1,63 @@
...
+  action        = "lambda:InvokeFunction"
+  function_name = aws_lambda_function.SyncApplicants.function_name
+  principal     = "events.amazonaws.com"
+  source_arn    = aws_cloudwatch_event_rule.hmbc_to_ien_masters.arn
+}
+
+
+# Scheduler to sync applicant and applicant-milestones
+resource "aws_cloudwatch_event_rule" "hmbc_to_ien_applicants" {
+  name                = "hmbc_to_ien_applicants-daily"
```
vyasworks, line 47: namespace pattern

terraform/lambdafunction.tf
```diff
@@ -0,0 +1,63 @@
...
+
+  environment {
+    variables = {}
+  }
+}
+
+
+# Scheduler to sync master tables
+resource "aws_cloudwatch_event_rule" "hmbc_to_ien_masters" {
+  name                = "hmbc_to_ien_masters-daily"
```
vyasworks, line 26: namespace pattern

vyasworks (commented):
Bunch of nit picks and standard practice corrections. Rest looks good

apps/api/src/syncdata.ts
```diff
@@ -0,0 +1,32 @@
+import { NestFactory } from '@nestjs/core';
+import { Context, Handler } from 'aws-lambda';
+import { AppModule } from './app.module';
+import { ExternalAPIService } from './applicant/external-api.service';
+import { AppLogger } from './common/logger.service';
+
+/**
```
vyasworks, line 7: :+1:

PR #66, opened 2022-04-20
add new endpoint, add conditional fields for new milestone

coryatwater-FW:
updated endpoint for retrieving a job based on job ID in order to update newly populated end date fields when a new milestone is added, also added conditional fields for offered position and candidate withdrew

### DIFF
--- a/apps/api/src/applicant/ienapplicant.controller.ts
+++ b/apps/api/src/applicant/ienapplicant.controller.ts
@@ -303,4 +303,34 @@ export class IENApplicantController {
       }
     }
   }
+
+  @ApiOperation({
+    summary: 'Get Specific Applicant job and related milestone',
+  })
+  @UseInterceptors(ClassSerializerInterceptor)
+  @ApiResponse({ status: HttpStatus.OK, type: EmptyResponse })
+  @HttpCode(HttpStatus.OK)
+  @RouteAcceptsRoles(
+    ValidRoles.HEALTH_AUTHORITY,
+    ValidRoles.HEALTH_MATCH,
+    ValidRoles.MINISTRY_OF_HEALTH,
+  )
+  @Get('/:id/job/:job_id')
+  getApplicantJob(
+    @Param('id') id: string,
+    @Param('job_id') job_id: string,
+  ): Promise<ApplicantJobRO | undefined> {
+    try {
+      return this.ienapplicantService.getApplicantJob(id, job_id);
+    } catch (e) {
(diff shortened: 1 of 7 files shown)


apps/api/src/applicant/ienapplicant.controller.ts
```diff
@@ -303,4 +303,34 @@ export class IENApplicantController {
...
+    try {
+      return this.ienapplicantService.getApplicantJob(id, job_id);
+    } catch (e) {
+      if (e instanceof NotFoundException) {
+        throw e;
+      } else if (e instanceof QueryFailedError) {
+        throw new BadRequestException(e);
+      } else {
+        // statements to handle any unspecified exceptions
+        throw new InternalServerErrorException('An unknown error occured while fetching ');
```
sunghwan-park-fw, line 332: In this way, it silences the original exception. How about logging it or including its message in a new exception?

apps/web/src/components/milestone-logs/recruitment/Milestone.tsx
```diff
@@ -11,44 +11,57 @@ import createValidator from 'class-validator-formik';
...
   jobId,
-  jobMilestones,
   setJobMilestones,
 }) => {
   const handleSubmit = async (values: any, { resetForm }: any) => {
     // hardcoding some values for now, specifically logged in user
     values.job_id = jobId.toString();
     values.added_by = '1';
 
+    if (values.status !== '304' && values.status !== '305') {
```
sunghwan-park-fw, line 52: why should the effective_date be cleared for other statuses?

sunghwan-park-fw (approved):
Just a suggestion and a question.

apps/web/src/components/milestone-logs/recruitment/Milestone.tsx
coryatwater-FW: effective_date is only show depending on those two Milestone options, otherwise it's undefined, I can't initialize a Formik Form with an undefined field, open to other solutions!

apps/api/src/applicant/ienapplicant.service.ts
```diff
@@ -224,6 +234,11 @@ export class IENApplicantService {
       data.reason = statusReason;
     }
 
+    data.reason_other = null;
```
sunghwan-park-fw, line 237: same as `data.reason_other = reason_other || null`?

sunghwan-park-fw (approved):

PR #68, opened 2022-04-20
Ien 208 update repo

dbayly-freshworks:
Updated out of date packages and regenerated the Yarn lock. This should address the vulnerabilities flagged by dependabot.


bhumin-fw (approved):
Looks good, 
Pr-check fail: might resolve by taking a pull from the main into your branch, not sure.
I did see nanoid: 3.3.3 in yarn.lock, in an open PR, added by Sunghwan(but no package.json update :) ). It might conflict, please try to make sure it won't override other details.

apps/web/package.json
```diff
@@ -26,7 +26,7 @@
     "cookie": "0.4.2",
     "formik": "2.2.9",
     "keycloak-js": "16.1.1",
-    "next": "12.0.4",
+    "next": "12.1.5",
```
sunghwan-park-fw, line 29: I tried this but it caused a lot of problems.

dbayly-freshworks: Yeah, this looks to be the cause of the broken tests. 2.1.4 seems to be stable.

PR #79, opened 2022-04-25
fix field text size, remove new reason button, fix table resizing whe…

coryatwater-FW:
fixed Field component text size, removed Add New Reason button, fixed ApplicantsTable from resizing when no data is present


dbayly-freshworks (approved):
Looks good

PR #84, opened 2022-04-26
IEN-250 added API to update user role

bhumin-fw:


sunghwan-park-fw (approved):

apps/api/src/employee/employee.service.ts
```diff
@@ -22,4 +24,48 @@ export class EmployeeService {
...
+      query.where = { name: ILike(`%${name}%`) };
+    }
+    return await this.employeeRepository.find(query);
+  }
+
+  /**
+   * Update role of the provided employee list
+   * @param ids Employee Ids whose role we are updating
+   * @param role role
+   */
```
dbayly-freshworks, line 49: Should we prevent an admin user from creating more admin users? I am just thinking about the risks of privilege escalation.

dbayly-freshworks (approved):
Looks good -- one comment

PR #96, opened 2022-04-28
IEN-273 | fix status of job competition and duration of milestone

sunghwan-park-fw:
- Duration - time passed since the last milestone
- Status - last milestone status appended by ‘Complete’ or ‘On Going’
- Sorting milestones - firstly by start_date, secondly by status id


package.json
```diff
@@ -19,7 +19,7 @@
     "watch": "run-p 'watch:common' 'watch:api' 'watch:web'",
     "start:api:local": "yarn workspace @ien/api start:dev:local",
     "start:web:local": "yarn workspace @ien/web start",
-    "start:local": "run-p 'watch:common' 'start:api:local' 'start:web:local'",
```
dbayly-freshworks, line 22: What was the issue here?

apps/web/src/components/milestone-logs/recruitment/Record.tsx
```diff
@@ -18,80 +18,62 @@ import { AddRecordModal } from '../../display/AddRecordModal';
...
 
+dayjs.extend(duration);
+dayjs.extend(relativeTime);
+
 interface RecordProps {
   job: ApplicantJobRO;
   update: (record?: ApplicantJobRO) => void;
 }
 
+const COMPLETED_STATUSES = [305, 306, 307, 308];
```
dbayly-freshworks, line 29: Will hard coded statuses cause issues here?

package.json
sunghwan-park-fw: `watch common` means that we are writing codes while keeping the app updated.
I thought `watch` command would be enough for the case and supposed that we run `start:local` after building all apps.

apps/web/src/components/milestone-logs/recruitment/Record.tsx
sunghwan-park-fw: > Will hard coded statuses cause issues here?

I don't think so, actually it's just refactored not added newly. btw, are you suggesting we should move them to database?

sunghwan-park-fw: I'll add an enum for the codes.

coryatwater-FW: we get those 4 values from the master-list.ts, I had added a to-do to fix the hardcoded values

bhumin-fw (approved):

PR #110, opened 2022-05-02
remove unused files from ehpr

sunghwan-park-fw:


bhumin-fw (approved):
It looks much cleaner now.

PR #111, opened 2022-05-03
IEN-287 enable added and updated by on status/milestone

bhumin-fw:
- Enable added and updated by on status/milestone
- Restrict setting user access. ROLEADMIN is only set on the database and not set using API.
- Fix small amount of code smell from the sonar cloud list

### DIFF
--- a/apps/api/src/common/auth.middleware.ts
+++ b/apps/api/src/common/auth.middleware.ts
@@ -1,11 +1,10 @@
 import { HttpException, HttpStatus, Inject, Injectable, NestMiddleware } from '@nestjs/common';
 import { NextFunction } from 'express';
 import { AuthService } from '../auth/auth.service';
-
 import { EmployeeService } from 'src/employee/employee.service';
-import { EmployeeEntity } from 'src/employee/employee.entity';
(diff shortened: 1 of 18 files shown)


apps/api/src/common/auth.middleware.ts
```diff
@@ -25,7 +25,7 @@ export class AuthenticationMiddleware implements NestMiddleware {
       throw new HttpException('Token not found', HttpStatus.BAD_GATEWAY);
     }
     try {
-      const user: any = await this.authService.getUserFromToken(token);
+      const user = await this.authService.getUserFromToken(token);
```
sunghwan-park-fw, line 28: Just a question though it's not about this PR.
It seems that both of auth guard and middleware validate the token and set the user. To separate concerns, can we make middleware validate token and guard validate role?

bhumin-fw: I feel we can move both into auth guard. As In reporting we might have roles that are separate from the existing role with very limited functionality.
AuthGuard will be a simple solution here, As if we do want to check partial URLs from the middleware then it will add an overhead for a developer to keep track of it, It needs to be updated for each URL which should check user token.

PR #129, opened 2022-05-10
show word Today for competition duration being added on current date

coryatwater-FW:
this seems more like a quick fix to me, I'm not sure at the moment what else to do with this.  I left a comment above the change, but it's basically comparing the start_date from the last milestone to the current date, excluding time values

### DIFF
--- a/apps/web/src/services/date-utils.ts
+++ b/apps/web/src/services/date-utils.ts
@@ -20,5 +20,13 @@ dayjs.extend(duration);
 dayjs.extend(relativeTime, { thresholds });
 
 export const getHumanizedDuration = (start: dayjs.ConfigType, end?: dayjs.ConfigType): string => {
+  // since we don't keep track of time, start_date from the DB gets set to 00:00:00:00
+  // so anything less than a day will take the current locale time and compare it to start_date
+  // ex if new status is added at 12pm locale time,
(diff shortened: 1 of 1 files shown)


sunghwan-park-fw:
> this seems more like a quick fix to me, I'm not sure at the moment what else to do with this. I left a comment above the change, but it's basically comparing the start_date from the last milestone to the current date, excluding time values

There seems to be no way to avoid confusion without time information. I wonder if we could use created_date if the start_date and created_date are the same date.

bhumin-fw:
> > this seems more like a quick fix to me, I'm not sure at the moment what else to do with this. I left a comment above the change, but it's basically comparing the start_date from the last milestone to the current date, excluding time values
> 
> There seems to be no way to avoid confusion without time information. I wonder if we could use created_date if the start_date and created_date are the same date.

@sunghwan-park-fw 
No, Created and Start_date both will be different in many statuses.

If the status date is today then we can display "today" if yesterday then we can display "yesterday"/"a day" and for more, the 2 days, simply put x days, a week, and so on.

It is simple, we won't show any of this like minutes, hours, etc.

coryatwater-FW:
> 

This is currently what the fix does

dbayly-freshworks (approved):
Looks ok to me!

PR #130, opened 2022-05-10
IEN-300 | seed applicants for cypress tests

sunghwan-park-fw:


sunghwan-park-fw:
> 

All data is generated by fakejs, so I believe they won't conflict with real-world data.
I didn't edit the insert query. Applicant data is created by backing up the table.

bhumin-fw:
> > 
> 
> All data is generated by fakejs, so I believe they won't conflict with real-world data. I didn't edit the insert query. Applicant data is created by backing up the table.

Sounds great, Please go ahead and merge it.

PR #136, opened 2022-05-12
IEN-298 | remove effective_date field when withdrawing

sunghwan-park-fw:


coryatwater-FW (approved):
should we remove from BE as well?

PR #151, opened 2022-05-20
IEN-345 | basic report page

sunghwan-park-fw:
- query report periods with sorting and pagination
- download function will be added later


apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -245,7 +245,6 @@ export class IENApplicantUtilService {
         // Let's not put hard limit right now, and accept multiple records here.
         if (previousStatus.length > 0) {
           const updateData = {
-            updated_by: data.added_by,
```
sunghwan-park-fw, line 248: It causes IEN-352 issue. It could be reasonable to update the previous one's updated_by because it updates end_date. But it might mislead users who don't know this background process, so I deleted it.

bhumin-fw: That is reasonable, Let's skip it.

bhumin-fw (approved):
This looks good,
We need to change the Reporting API for applicants as per the spreadsheet.
Please go ahead and merge it, I will handle API changes in the next PR.

PR #156, opened 2022-05-25
change role

sunghwan-park-fw:


apps/web/src/components/Keycloak.tsx
```diff
@@ -23,6 +23,7 @@ const withAuth = (Component: React.FunctionComponent, roles: ValidRoles[]) => {
       if (!authUser && !authUserLoading && kc.initialized && !kc?.keycloak?.authenticated) {
         router.replace('/login');
       }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
dbayly-freshworks, line 26: We might want to create a tech debt ticket to look into exhaustive deps.

apps/web/src/components/display/AddRecordModal.tsx
```diff
@@ -87,7 +87,7 @@ export const AddRecordModal: React.FC<AddRecordProps> = (props: AddRecordProps)
   };
 
   return (
-    <Modal open={visible} handleClose={close}>
+    <Modal open={visible} handleClose={() => void 0}>
```
dbayly-freshworks, line 90: I'm curious about whats going on here. How does this function work?

dbayly-freshworks (approved):
This all looks good to me, I just have two small questions. I am really glad ValidRoles is getting shifted to common.

bhumin-fw (approved):
It looks good

apps/web/src/components/display/AddRecordModal.tsx
sunghwan-park-fw: it's same as `undefined`. Modal requires handleClose prop but it isn't called ever and I don't know why.

PR #157, opened 2022-05-25
IEN-345 | download report with sheet 1 - registered applicants

sunghwan-park-fw:


bhumin-fw (approved):
@sunghwan-park-fw 
This looks good.

I have one point to discuss, Let's take it offline and discuss it today after standup.
The point is: When clicking the download(button for a specific period) data(in CSV or Excel), We need to fetch new data from the API and pass "from" (for now April 1, 2022) and "to" (Selected periods' end-date) query params.

PR #162, opened 2022-05-27
IEN-331 Communities/Job Locations list updated

bhumin-fw:


apps/api/src/migration/1653542688339-seedlocationdata.ts
```diff
@@ -0,0 +1,103 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class Seedlocationdata1653542688334 implements MigrationInterface {
+  name = 'seedlocationdata1653542688334';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DELETE FROM public.ien_applicant_jobs_job_location_ien_job_locations WHERE ien_job_locations_id > 84`,
```
sunghwan-park-fw, line 8: is this `ien_applicant_jobs_job_location_ien_job_locations` correct?

apps/api/src/migration/1653542688339-seedlocationdata.ts
```diff
@@ -0,0 +1,103 @@
...
+    );
+    await queryRunner.query(`DELETE FROM public.ien_job_locations WHERE id > 84`);
```
sunghwan-park-fw, line 10: to run the next insert queries, don't we need to delete all rows?

apps/api/src/migration/1653542688339-seedlocationdata.ts
bhumin-fw: Yes, It is a job-location mapping table(auto-generated when we allow multiple job location selections).
Here we are updating the new location and the max id will be 84. If id>84, those records are not removed then it won't allow us to delete it.
So removing any extra locations that is attached earlier

sunghwan-park-fw (approved):

apps/api/src/migration/1653542688339-seedlocationdata.ts
bhumin-fw: We already have data in DEV and QA associated with a few locations.
Here we run UPSERT with new data(INSERT or UPDATE).

It will help us to preserve already selected locations for existing jobs. For DEV and TEST, these are simple solutions that we can apply.

Let's Say, If we already have data in PROD, Then we have to write a query that updated existing data with a new id, Ald report removed locations that are already selected for any job.
Here most of the time first we do identify all such data and take decisions on the management side on what to do with those data.

PR #163, opened 2022-05-27
IEN-351 | fix applicant filter

sunghwan-park-fw:
- prevent redundant queries caused by react state updates
- fix loading spinner for applicants, users, and reports table


bhumin-fw (approved):
Looks good, multiple changes makes more cleaner implementation (y)

PR #172, opened 2022-06-01
IEN-343 Upgrade postgres version 13.4 to 13.6

bhumin-fw:


bhumin-fw:
@vyasworks 
It seems working, Upgrade from the console and putting a new version in TF looks good.

It detected that the DB version upgrades from 13.4 to 13.6 but didn't include that changes in the final plan.
**Plan: 0 to add, 2 to change, 0 to destroy.**
Here **2 to change** consider 2 lambda functions.

bhumin-fw:
@vyasworks 
Merging it for now. We will chat if we have to revert it. As it block other deployment.

PR #173, opened 2022-06-01
IEN-346 Added Audit log to help Developer to track down data flow

bhumin-fw:


apps/api/src/applicant/ienapplicant.controller.ts
```diff
@@ -59,6 +59,8 @@ export class IENApplicantController {
   @Get('/')
   async getApplicants(@Query() filter: IENApplicantFilterAPIDTO): Promise<[ApplicantRO[], number]> {
     try {
+      this.logger.log(`list applicant called with the below filter parameters`);
```
sunghwan-park-fw, line 62: Logging filter might be good for debugging but not provide any useful information for audit. Also sometimes consecutive loggings are apart from each other when they are actually written to a file.

apps/api/src/applicant/ienapplicant.controller.ts
```diff
@@ -221,6 +232,8 @@ export class IENApplicantController {
     @Body() jobData: IENApplicantJobCreateUpdateAPIDTO,
   ): Promise<ApplicantJobRO | undefined> {
     try {
+      this.logger.log(`Add job competition for the applicant ${id}`);
+      this.logger.log({ jobData });
```
sunghwan-park-fw, line 236: job data has recruiter name. could this be exposing personal information?

apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -110,6 +110,10 @@ export class IENApplicantUtilService {
 
     if (status) {
       const status_list = await this.fetchChildStatusList(status);
+      this.logger.log(
```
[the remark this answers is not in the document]
bhumin-fw: Added it as we are getting a list of child milestone list, So that help to verify it in debug what all milestone is checked.

apps/api/src/applicant/ienapplicant.controller.ts
```diff
@@ -304,6 +319,8 @@ export class IENApplicantController {
     @Query() options: IENApplicantJobQueryDTO,
   ): Promise<[ApplicantJobRO[], number]> {
     try {
+      this.logger.log(`Fetch job competition for the applicant ${id} with below options/filter`);
+      this.logger.log({});
```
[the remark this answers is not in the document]
bhumin-fw: It was added by mistake, let me remove it.

apps/api/src/applicant/ienapplicant.controller.ts
bhumin-fw: agree with your point, What do you suggest. is removing filter param fine?

sunghwan-park-fw (approved):
I left some comments.

PR #180, opened 2022-06-03
IEN-384 delete applicant milestone

bhumin-fw:


apps/api/src/applicant/ienapplicant.controller.ts
```diff
@@ -192,6 +193,28 @@ export class IENApplicantController {
     }
   }
 
+  @ApiOperation({
+    summary: 'Delete applicant milestone/status',
+  })
+  @UseInterceptors(ClassSerializerInterceptor)
+  @Delete('/:id/status/:status_id')
+  @RouteAcceptsRoles(ValidRoles.HEALTH_AUTHORITY, ValidRoles.HEALTH_MATCH)
```
coryatwater-FW, line 201: the story only mentions HMBC staff, are we including HA now too?

coryatwater-FW (approved):
looks good to me, just one general question

apps/api/src/applicant/ienapplicant.controller.ts
bhumin-fw: Seems an interesting point to discuss here.
Actually, that was there for all other APIs. We are not allowing superuser/roleadmin to enable HA role for any user as of now. So that won't affect anything. When we do enable HA in the system we must have to add an additional filter apart from the role so that HA can see only those applications which are referred to that HA.

In my opinion, we can keep HA enable for all APIs for now.
What is your thought here @coryatwater-FW @sunghwan-park-fw ?

coryatwater-FW: hmm, do we know if HA will have the ability to delete though?

I guess it doesn't really matter as we just either remove HA from routes when enabled and not required, or add them to routes when enabled and required

bhumin-fw: Yes, HA can add or delete milestone in recruitment. That’s their main activity here in IEN app.

sunghwan-park-fw (approved):

PR #184, opened 2022-06-06
add ha recruiter name to job record

coryatwater-FW:


coryatwater-FW:
quick question, I know HA recruiter name is required, so do we even need a 'N/A' check?

bhumin-fw:
> @bhumin-fw quick question, I know HA recruiter name is required, so do we even need a 'N/A' check?

@coryatwater-FW No, Sometimes we have records but the recruiter may not work with this firm anymore. So they do add N/A and later update it when assigning a new recruiter.

dbayly-freshworks (approved):

PR #186, opened 2022-06-06
add duration column value to milestone table

sunghwan-park-fw:
- show duration of each milestone in the milestone table
- increase github action timeout for e2e+a11y up to 20 minutes

### DIFF
--- a/.github/workflows/pr-check-e2e.yml
+++ b/.github/workflows/pr-check-e2e.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   pr-checks:
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
     environment:
       name: dev
     steps:
--- a/apps/web/src/components/milestone-logs/MilestoneTable.tsx
+++ b/apps/web/src/components/milestone-logs/MilestoneTable.tsx
@@ -74,15 +74,23 @@ export const MilestoneTable = ({ parentStatus }: MilestoneTableProps) => {
 
   return (
     <div>
-      <div className='opacity-50 pb-2'>Showing {milestones.length} logs</div>
+      <div className='opacity-60 pb-2'>Showing {milestones.length} logs</div>
       <div>
         <table className='w-full'>
           <thead className=''>
             <tr className='bg-bcLightGray text-bcDeepBlack text-sm text-left border-b-2 border-yellow-300'>
-              <th className='py-4 pl-8'>Milestones</th>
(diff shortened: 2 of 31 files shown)


bhumin-fw (approved):

apps/web/src/components/milestone-logs/MilestoneTable.tsx
```diff
@@ -41,6 +42,18 @@ export const MilestoneTable = ({ milestones }: MilestoneTableProps) => {
     );
   };
 
+  const getDuration = (milestone: ApplicantStatusAuditRO): string => {
+    const start = milestone.start_date;
+    let end = milestone.end_date;
+    if (!end) {
+      const nextMilestone = milestones.filter(m => m.status.id > milestone.status.id)[0];
```
bhumin-fw, line 49: We should sort based on the date and get the next milestone date here.
- We are hoping that ID is in incremental order and,
- assuming that no one gets the latest date on any previous milestone.
If both of the above cases are in place this implemented logic will work fine.

Approved it as per MVP requirements.

bhumin-fw (approved):

dbayly-freshworks (approved):
One comment, otherwise good.

.github/workflows/pr-check-e2e.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   pr-checks:
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
```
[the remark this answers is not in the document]
sunghwan-park-fw: Cypress tests cover almost every function, so it takes some time. However, the login process impacts heavily because keycloak sometimes responds in dozens of seconds. I have thought about a way of removing login step only for the testing, like disabling auth on backend, but I want to do it as a last resort.

apps/web/src/components/milestone-logs/MilestoneTable.tsx
sunghwan-park-fw: I noticed that `milestones` are filtered by the selected tab and am going to fix it.
MelestoneTable is just for milestones except ones under recruitment category and they should come from ATS. So condition of `assuming that no one gets the latest date on any previous milestone` shouldn't be on us. With this condition, I am going to sort milestones by `start_date.`

sunghwan-park-fw:
@dbayly-freshworks 
Total time  has been reduced by removing logout steps using cypress session, but still exceeds 10 minutes.

PR #192, opened 2022-06-08
IEN-422 first version of apis for reports 3 to 7

bhumin-fw:
- Added API for reports 3 to 7
- APIs with fixed formate and partially confirmed Queries data

### DIFF
--- a/apps/api/src/report/report.controller.ts
+++ b/apps/api/src/report/report.controller.ts
@@ -1,10 +1,9 @@
-import { Controller, Get, HttpStatus, Inject, Logger, Query, UseGuards } from '@nestjs/common';
-import { ApiOperation, ApiResponse, ApiTags } from '@nestjs/swagger';
+import { Controller, Get, Inject, Logger, Query, UseGuards } from '@nestjs/common';
+import { ApiOperation, ApiTags } from '@nestjs/swagger';
 import { ValidRoles } from '@ien/common';
 import { AuthGuard } from 'src/auth/auth.guard';
 import { RouteAcceptsRoles } from 'src/common/decorators';
 import { ReportService } from './report.service';
-import { PeriodRO } from '../common/ro/period.ro';
 import { AppLogger } from 'src/common/logger.service';
 
 @Controller('reports')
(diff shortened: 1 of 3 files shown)


apps/api/src/report/report.controller.ts
```diff
@@ -47,4 +46,81 @@ export class ReportController {
...
+    @Query('from') from: string,
+    @Query('to') to: string,
+  ): Promise<object[]> {
+    this.logger.log(
+      `Report: Applicants count of Active, Withdrawn and Hired (from ${from} to ${to})`,
+    );
+    return this.reportService.getHiredWithdrawnActiveCount(from, to);
+  }
+
+  @ApiOperation({ summary: 'Report 4: hired-withdrawn-active applicants' })
```
coryatwater-FW, line 67: summary should be licensing stage applicants?

coryatwater-FW (approved):
looks good to me

apps/api/src/report/report.util.service.ts
```diff
@@ -119,6 +119,297 @@ export class ReportUtilService {
...
+          t1.id,
+          COALESCE((SELECT max(sa.status_id) FROM public.ien_applicant_status_audit sa 
+            WHERE sa.applicant_id=t1.id AND sa.status_id IN (19, 16)), 0) AS has_license
+        FROM (
+          SELECT 
+            applicants.id,
+            CASE WHEN COALESCE(ien_staus_hired.status_id, 0) > 0 THEN 1 ELSE 0 END as hired,
+            CASE WHEN COALESCE(ien_staus_withdrew.status_id, 0) > 0 THEN 1 ELSE 0 END as withdrawn
+          FROM public.ien_applicants as applicants
+          LEFT JOIN public.ien_applicant_status_audit ien_staus_hired
```
sunghwan-park-fw, line 184: Isn't it same one join with two conditions combined by 'or'?

bhumin-fw: @sunghwan-park-fw
No, its not same as you may receive both status in some case. In such case we have to run group by and give priority to withdrew first in selection.

Use 2 instance of same table to find each detail separately help us to keep only one raw for an applicant. And we do not need to run a process of group by, partitions and provide row number based on order withdrawal then hired and pic row 0 for the same.

sunghwan-park-fw (approved):

PR #203, opened 2022-06-13
remove 2 more major smells

coryatwater-FW:


sunghwan-park-fw (approved):
I wonder why sonarcloud pr check have reported 0 code smell so far.

PR #206, opened 2022-06-14
Added log for request-response with the help of middleware.

bhumin-fw:


bhumin-fw:
@sunghwan-park-fw 
Added a middleware that log request and relative response. It will log the below data from the request & response object combined.
`${method} ${originalUrl} ${statusCode} - ${userAgent} ${ip}`
cc: @coryatwater-FW 

Please provide your thought on this.
If this looks good, I will remove the informative message/log from all the controller functions.
And keep only business or logical messages in service, util & common files.

sunghwan-park-fw:
Looks good. I just wonder if we need to log requests before processing apart from logging response.

bhumin-fw:
> Looks good. I just wonder if we need to log requests before processing apart from logging response.

@sunghwan-park-fw 
I don’t think so, as i have added log under close loop.
We will only miss request log if sever suddenly stopped working in between request and response.

Do you think that’s a high severity information we might miss?

sunghwan-park-fw:
> > Looks good. I just wonder if we need to log requests before processing apart from logging response.
> 
> @sunghwan-park-fw I don’t think so, as i have added log under close loop. We will only miss request log if sever suddenly stopped working in between request and response.
> 
> Do you think that’s a high severity information we might miss?

No. It seems enough as it is.

sunghwan-park-fw (approved):

PR #212, opened 2022-06-17
IEN-443 Audit Sync Applicants run

bhumin-fw:
- IEN-443 Add table to log audit for applicant sync activity
- Add an end-point to fetch the last successful Audit run.
- Fix 5 Sonar MAINTAINABILITY points

### DIFF
--- a/apps/api/src/applicant/external-api.service.ts
+++ b/apps/api/src/applicant/external-api.service.ts
@@ -2,17 +2,14 @@
 import { Inject, Injectable, Logger } from '@nestjs/common';
 import { ExternalRequest } from 'src/common/external-request';
 import { InjectRepository } from '@nestjs/typeorm';
-import { IENHaPcn } from './entity/ienhapcn.entity';
 import { In, Repository } from 'typeorm';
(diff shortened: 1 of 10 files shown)


apps/api/src/applicant/external-api.service.ts
```diff
@@ -462,36 +495,56 @@ export class ExternalAPIService {
...
   }
+
+  /** create applicant-milestone object */
+  mapMilestoneData(
+    allowedMilestones: string | any[],
+    m: {
+      id: any;
+      start_date: any;
+      created_date: any;
+      note: any;
```
coryatwater-FW, line 513: how come 'any'?

coryatwater-FW (approved):
looks good to me, just a question out of curiosity

apps/api/src/applicant/external-api.service.ts
bhumin-fw: Auto ts lint, picks any if it doesn’t know its source type. In this case it’s coming from ATS and read as a json object. I can fix few any, let me do that.

sunghwan-park-fw (approved):

PR #214, opened 2022-06-17
IEN-432 | format reports

sunghwan-park-fw:


bhumin-fw (approved):
This looks good to me.
Some Sonar code smell & duplicate line indication(that generate for 2 block which has identical key and functions).

**possible fix/solution you can apply:**
code-smell: Try to move the nested block of IF conditions into a separate function.
Duplicate code: Try to change the col value by +/- 1 for a few columns. The header function looks similar you can handle it as a separate function too.

Note: These suggestions make no sense to me, And if anyone raises any query from the client-side based on this report, we can revert with a proper explanation. So do not spend extra time on this, if you find quick adjustment help then do it, else leave as it is. :) (y)

PR #220, opened 2022-06-23
IEN 458 update reports having hold status

bhumin-fw:


apps/api/src/report/report.util.service.ts
```diff
@@ -120,44 +120,113 @@ export class ReportUtilService {
...
+        FROM public.ien_applicant_status_audit 
+        WHERE status_id=20 
+          AND start_date::date <= '${to}'
+        GROUP BY applicant_id
+      ), reactive_applicants AS (
+        SELECT wa.applicant_id, min(ien.start_date) as start_date
+        FROM withdrawn_applicants wa INNER JOIN public.ien_applicant_status_audit ien ON wa.applicant_id=ien.applicant_id
+        WHERE ien.start_date::date >= wa.start_date
+          AND ien.start_date::date <= '${to}'
+          AND ien.status_id != 20
```
sunghwan-park-fw, line 160: Is the reason that it has status withdrew(20) even though it has later milestone than 20,  because status comes from ATS?

bhumin-fw: @sunghwan-park-fw 
Yes, applicants can restart application too even after withdrawal.
If applicant is not responsive for longer period then it will be withdrawn but later applicant came again and want to start from where it left.

bhumin-fw: Also applicant can withdraw from recruitment or immigration stage too. Or hold application by withdrawing it and start again.

sunghwan-park-fw (approved):
It will take days to digest the queries.
I'd have done it when you first wrote them.
Anyway, I don't want to block you.
I hope to have some time to write some documents dissecting and analyzing them later.

Just one suggestion, how about adding descriptions in the summary sheet to the top of the comment block?

PR #221, opened 2022-06-23
bug fixes for ien 434

coryatwater-FW:
IEN-460 - prevent download of empty files and added toast to display reason
IEN-459, 462, 461 - added validation and error messages for when a user manually types in dates


bhumin-fw (approved):
This looks good, Please change the min-date to 2001-01-01 for now, later we will update it by consulting with HMBC/MOH.

PR #223, opened 2022-06-23
add dto for time period for data extract

coryatwater-FW:
added dto to make start and end date required for data extract


bhumin-fw (approved):
Looks good, This seems straight forword.

PR #229, opened 2022-06-27
IEN-457 standard logic for report 1 to 9

bhumin-fw:


coryatwater-FW (approved):
looks good to me, it'll obviously take more than a single read

PR #238, opened 2022-07-01
IEN-495 | user can have multiple roles

sunghwan-park-fw:


apps/api/src/employee/employee.service.ts
```diff
@@ -58,16 +56,14 @@ export class EmployeeService {
...
-    return getManager()
-      .createQueryBuilder(EmployeeEntity, 'employee')
-      .select('employee.*')
-      .addSelect('users.id', 'user_id')
-      .leftJoin('ien_users', 'users', 'employee.email = users.email')
-      .where('employee.keycloak_id = :keyclock', { keyclock: keycloakId }) // WHERE t3.event = 2019
-      .getRawOne();
+    const employee = await this.employeeRepository.findOne({ where: { keycloakId } });
+    if (employee) {
+      const user = await this.getUser(employee.email);
```
bhumin-fw, line 61: It will send 2 query request and network call. Use of  join query will complete the same in one network call and request. Typeorm find() and other function does not allow us to join out of relations. https://typeorm.io/find-options based on this reference I have added query builder that generate same query but where clause can be manually update and does not required any FK or index on it.


Please let me know your concern or query here? let's identify best way for the same so we can use it further(we have to fetch HA based on organization name too).

sunghwan-park-fw: I didn't like those two separate queries, but it doesn't populate roles properly. If there is a way to resolve roles, please let me know.

bhumin-fw (approved):
This seems good now for the first step in user management.

PR #241, opened 2022-07-07
IEN-5 | Add Applicant

coryatwater-FW:


sunghwan-park-fw:
@coryatwater-FW 
I made some changes and fixed sonar error of duplicate codes.
btw, added applicant doesn't have applicant_id value that is displayed in the first column of applicant list table.

coryatwater-FW:
> 

awesome!  Yes that's the intention.  applicant_id is a reference id from the ATS system, so when we add a new IEN in the application it will be null

apps/api/src/applicant/dto/ienapplicant-create.dto.ts
```diff
@@ -84,8 +91,12 @@ export class IENApplicantCreateUpdateAPIDTO extends IENApplicantCreateUpdateDTO
     ],
   })
   @IsArray()
-  @IsOptional()
-  nursing_educations?: JSON;
+  @ArrayMinSize(1, { message: 'At least 1 Education is required' })
+  @ArrayNotEmpty({ message: 'Education is required' })
+  @ValidateIf(e => e.nursing_educations.length > 0)
+  @ValidateNested()
+  @Type(() => NursingEducationDTO)
```
sunghwan-park-fw, line 97: Swagger doc doesn't show the schema for nested nursing education dto.
We might need to define separate dto in api project. It means sharing concept of common project doesn't work for validators and swagger api.

sunghwan-park-fw (approved):

PR #243, opened 2022-07-12
IEN-498, IEN-500 | assign roles

sunghwan-park-fw:
- user details page
- assign roles
- revoke access


sunghwan-park-fw:
Sorry for a too big pull request.
It's because role and acl affects almost every codes.

apps/api/src/common/auth.middleware.ts
```diff
@@ -29,10 +28,10 @@ export class AuthenticationMiddleware implements NestMiddleware {
...
 
       // Retrieve user roles
-      const resourceAccess = user['resource_access'];
-      if (resourceAccess && process.env.AUTH_CLIENTID) {
-        res.locals.roles = resourceAccess[process.env.AUTH_CLIENTID]?.roles;
-      }
+      // const resourceAccess = user['resource_access'];
+      // if (resourceAccess && process.env.AUTH_CLIENTID) {
+      //   res.locals.roles = resourceAccess[process.env.AUTH_CLIENTID]?.roles;
+      // }
```
dbayly-freshworks, line 34: Commented out code here

apps/api/src/migration/1657325456713-MapRoleAcl.ts
```diff
@@ -0,0 +1,210 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class MapRoleAcl1657325456713 implements MigrationInterface {
```
dbayly-freshworks, line 3: If possible, it would be best to use the query builder to create these migrations instead just using .query.

apps/api/src/migration/1657325456713-MapRoleAcl.ts
```diff
@@ -0,0 +1,210 @@
...
+              (SELECT id FROM "role" WHERE slug = 'reporting'),
+              u.id
+            );
+          
+        END LOOP;
+      END; $$;
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
```
dbayly-freshworks, line 176: Have you tested that the down function works as expected?

apps/api/src/employee/employee.controller.ts
```diff
@@ -28,9 +30,15 @@ export class EmployeeController {
...
-  async getEmployee(@Req() req: RequestObj): Promise<EmployeeEntity> {
-    return req.user;
+  @Get()
+  async getEmployee(@Req() req: RequestObj, @Query('id') id: string): Promise<EmployeeRO> {
+    if (!id) return req.user;
+
+    const employee = await this.employeeService.getEmployee(id);
+    if (!employee) {
+      throw new NotFoundException(`User with id '${id}' not found`);
+    }
```
dbayly-freshworks, line 40: This route may need a bit of a refactor. The intention behind it was to have a route to get the user object of the logged in user. All the work is done by the guard and it should be returned. 

It looks like you're adding a new purpose to the route, eg getting another user's user object. We might want to split this into another route. We should also be careful about what information we send back to a user who is not the logged in user.

dbayly-freshworks (commented):
This is a little too large to review in one go, but here are my initial comments.

apps/api/src/migration/1657325456713-MapRoleAcl.ts
bhumin-fw: In my opinion, based on typeorm document. Using query is a more flexible way to play around.
Query builder is a way to use migration API. And that won't work in all the cases with generate statements(like if-exist partial unique constrain or migrate data from one form to another form).
but query works well in all the cases, by simply running `typeorm migration: generate -n <name>`. Moreover, sometimes it requires manual crossover like changing a column from array/JSON to many-to-many relation without data loss, this scenario handles pretty nicely with query.

apps/api/src/employee/employee.controller.ts
bhumin-fw: Agree with David's point here.
We can easily control with auth guard, only the authorized(super user) user to reach this route controller.

apps/api/src/migration/1657325456713-MapRoleAcl.ts
sunghwan-park-fw: In some cases, query builder gets more complex than raw queries or it takes too much time. I need to study more about query builder. Anyway, I wanted to keep it consistent, so I chose to stick to raw query than mixing with query builder.
I tried to generate  a migration and it returned a bunch of queries not related to my work. Next time, I'll try and pick just what I need from the generated.

apps/api/src/common/auth.middleware.ts
```diff
@@ -29,10 +28,10 @@ export class AuthenticationMiddleware implements NestMiddleware {
       res.locals.kcUser = user;
 
       // Retrieve user roles
-      const resourceAccess = user['resource_access'];
```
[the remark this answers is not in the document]
sunghwan-park-fw: Deleted

apps/api/src/common/auth.middleware.ts
sunghwan-park-fw: Deleted

bhumin-fw (approved):

PR #249, opened 2022-07-19
add milestone outside recruitment

coryatwater-FW:


apps/api/src/applicant/ienapplicant.service.ts
```diff
@@ -239,12 +244,12 @@ export class IENApplicantService {
...
     }
-    if (!job) {
+    if (data.status.parent?.id === 10003 && !job) {
       throw new BadRequestException(`Competition/job are required to add a milestone`);
     }
 
-    if (user_id) {
-      const added_by_data = await this.ienUsersRepository.findOne(user_id);
+    if (user.user_id) {
+      const added_by_data = await this.ienUsersRepository.findOne(user.user_id);
```
sunghwan-park-fw, line 252: Can we just set user_id? `data.added_by = user.user_id` instead of querying user.

apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -176,12 +177,15 @@ export class IENApplicantUtilService {
...
-        throw new BadRequestException(
-          `Only recruitment-related milestones/statuses are allowed here`,
-        );
-      }
+
+    if (user.roles.some(r => r.slug === RoleSlug.Admin)) {
+      return statusObj;
+    }
+
+    if (statusObj && !user.ha_pcn_id && statusObj.parent?.id != 10003) {
```
sunghwan-park-fw, line 185: We'd better move this logic to caller because `getStatusById` should be the atomic method returning just status by id.

apps/web/src/components/milestone-logs/MilestoneTable.tsx
```diff
@@ -1,20 +1,49 @@
...
+import editIcon from '@assets/img/edit.svg';
 
 interface MilestoneTableProps {
   parentStatus: number;
 }
 
 const DEFAULT_TAB_PAGE_SIZE = 5;
 
+const wasAddedByLoggedInUser = (loggedInId?: string | null, addedById?: string) => {
+  return loggedInId && loggedInId.toString() === addedById;
```
sunghwan-park-fw, line 19: Why do we need `toString()`?

apps/web/src/components/milestone-logs/recruitment/Milestone.tsx
```diff
@@ -120,60 +150,68 @@ export const EditMilestone: React.FC<EditMilestoneProps> = props => {
...
-                disabled={!!editing && milestone === editing}
-              >
-                <img src={editIcon.src} alt='edit milestone' />
-              </button>
-              {deleteButton()}
-            </div>
-            <span className='text-sm text-black break-words'>
-              {milestone.notes || 'No Notes Added'}
+
+  const isRecruitmentRelated = () => {
```
sunghwan-park-fw, line 154: method name hints it will return boolean.

apps/web/src/services/master-lists.ts
```diff
@@ -31,10 +31,15 @@ export interface MilestoneType extends StyleOption {
 }
 
 // milestone status' for adding milestones
-export const useGetMilestoneOptions = (): MilestoneType[] => {
+export const useGetMilestoneOptions = (statusId: number): MilestoneType[] => {
   const { data: milestones } = useSWRImmutable('ienmaster/status', fetcher);
 
-  return milestones?.data.filter((item: { id: number }) => item.id == 10003)[0].children;
+  // All tab defaults to 10000,
```
sunghwan-park-fw, line 37: what does this comment mean?
I wonder if there is a case that `statusId` could be 10000.

sunghwan-park-fw (commented):
I left some comments.

apps/api/src/applicant/ienapplicant.service.ts
coryatwater-FW: the query returns a User object containing name, email etc.  not just assigning the id to the field

apps/web/src/components/milestone-logs/MilestoneTable.tsx
coryatwater-FW: good question, I think originally I wrote a similar one and loggedInId was a type Number at the time, will fix this mistake!

apps/web/src/components/milestone-logs/recruitment/Milestone.tsx
coryatwater-FW: yep good catch, will change fn name

apps/web/src/services/master-lists.ts
coryatwater-FW: when a new applicant gets created, they have no 'default' status for the details page to go to, so it defaults to 10000 which is the index value of the 'All' tab in the Applicants Table page, this just prevents it from erroring out.  I suppose we could just change the default landing tab to Intake if theres not status'

apps/api/src/applicant/ienapplicant.service.ts
sunghwan-park-fw: it doesn't return `data` object but passes it to addApplicantStatusAudit method. My understanding is that we don't need to resolve relation when adding an entry but addApplicantStatusAudit should do.

sunghwan-park-fw (approved):

PR #252, opened 2022-07-21
IEN - 512 Job ID is nolonger mandatory

dbayly-freshworks:
Removed the `not empty` flag for Job Id 
Added UUID on before insert 
Also tweeked package log to accept more recent versions of node. My docker got wiped out and the current version of Node-alpine -16 will have Node v 16.16.0 installed (which will cause the application to exit because it's not 16.15.1.

### DIFF
--- a/apps/api/src/applicant/entity/ienjob.entity.ts
+++ b/apps/api/src/applicant/entity/ienjob.entity.ts
@@ -1,4 +1,5 @@
 import { Exclude } from 'class-transformer';
+import { randomUUID } from 'crypto';
 import sortStatus from 'src/common/util';
 import {
   Entity,
@@ -11,6 +12,7 @@ import {
   AfterLoad,
   ManyToMany,
   JoinTable,
+  BeforeInsert,
 } from 'typeorm';
 import { IENApplicantStatusAudit } from './ienapplicant-status-audit.entity';
 import { IENApplicant } from './ienapplicant.entity';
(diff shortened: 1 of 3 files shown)


bhumin-fw (approved):

apps/api/src/applicant/entity/ienjob.entity.ts
```diff
@@ -66,4 +68,10 @@ export class IENApplicantJob {
       this.status_audit = sortStatus(this.status_audit);
     }
   }
+  @BeforeInsert()
+  beforeInsert() {
```
sunghwan-park-fw, line 72: this job_id is for reference to a real job posting. so I don't think it makes sense to generate a random id.

dbayly-freshworks: I think having this field blank and not generating one will create more issues than it solves. 
If we don't have it, we'll have trouble differentiating between job competitions. Also, we'll have an awkward edge case when checking for the uniqueness of the ID field. 

Plus we'll need to sync jobs with ATS eventually and having that uuid will allow that process to go smoothly. 

The main case I can against it is in the case of de-duplication or matching with an existing job within ATS. I don't think that's within the scope of our work.

sunghwan-park-fw: I get your point. Another question. randomUUID is a little long and verbose to display. Do we want to abbreviate it, like the first 4 bytes?

sunghwan-park-fw (approved):

apps/api/src/applicant/entity/ienjob.entity.ts
dbayly-freshworks: That works 💯

PR #256, opened 2022-07-22
IEN-533 | fix default milestone tab

sunghwan-park-fw:


.github/workflows/pr-check-e2e.yml
```diff
@@ -28,8 +28,8 @@ jobs:
       - name: Install dependencies
         run: yarn install
 
-      - name: Run application and cypress/accessibility tests
```
sunghwan-park-fw, line 31: Now, we can split pa11y tests from cypress tests because seeding data doesn't depend on cypress tests.

sunghwan-park-fw:
API integration tests fail randomly because of database connection failure, but why?

PR #282, opened 2022-08-03
stub out user sync to ats

coryatwater-FW:
does not include API key header as of yet

### DIFF
--- a/apps/api/src/applicant/external-api.controller.ts
+++ b/apps/api/src/applicant/external-api.controller.ts
@@ -93,10 +93,8 @@ export class ExternalAPIController {
   })
   @UseInterceptors(ClassSerializerInterceptor)
   @HttpCode(HttpStatus.OK)
-  @Get('/sync-users')
-  async syncUsersWithATS(
-    @Query() filter: IENUserFilterAPIDTO,
-  ): Promise<[data: IENUsers[], count: number]> {
-    return this.externalAPIService.syncUsers(filter);
+  @Get('/users')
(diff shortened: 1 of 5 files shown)


apps/api/src/applicant/external-api.controller.ts
```diff
@@ -84,4 +87,16 @@ export class ExternalAPIController {
   async getLatestSuccessfulSync(): Promise<SyncApplicantsAudit[]> {
     return this.externalAPIService.getLatestSuccessfulSync();
   }
+
+  @ApiOperation({
+    summary: `Sync Users to ATS`,
+  })
+  @UseInterceptors(ClassSerializerInterceptor)
+  @HttpCode(HttpStatus.OK)
+  @Get('/sync-users')
```
sunghwan-park-fw, line 96: This naming sync might hint that it will update data.
How about path as /users and method as getUsers

packages/common/src/enum/organizations.ts
```diff
@@ -24,3 +24,16 @@ export const EmailDomains = {
   'northernhealth.ca': Organizations.NORTHERN_HEALTH,
   'vch.ca': Organizations.VANCOUVER_COASTAL_HEALTH,
 };
+
+export const EmailDomainByAcronym = {
```
sunghwan-park-fw, line 28: Constants related to organization could be refactored but no idea at the moment.

sunghwan-park-fw (approved):

packages/common/src/enum/organizations.ts
coryatwater-FW: agreed

PR #286, opened 2022-08-04
only show and allow adding job competitions from same HA

coryatwater-FW:


apps/web/src/components/display/AddRecordModal.tsx
```diff
@@ -100,21 +102,34 @@ export const AddRecordModal: React.FC<AddRecordProps> = (props: AddRecordProps)
...
-                        value={haPcn?.data?.find(s => s.id == field.value)}
-                        onBlur={field.onBlur}
-                        onChange={value => form.setFieldValue(field.name, `${value?.id}`)}
-                        options={haPcn?.data?.map(s => ({ ...s, isDisabled: s.id == field.value }))}
-                        getOptionLabel={option => `${option.title}`}
-                        styles={getSelectStyleOverride<RecordTypeOptions>()}
-                      />
-                    )}
-                  />
+                  {authUser?.ha_pcn_id ? (
```
coryatwater-FW, line 105: I suppose this would be temporary until hmbc

apps/web/src/components/display/AddRecordModal.tsx
```diff
@@ -100,21 +102,34 @@ export const AddRecordModal: React.FC<AddRecordProps> = (props: AddRecordProps)
...
+                    <Field
```
sunghwan-park-fw, line 106: How about adding `isDisabled` prop to Select component?

apps/web/src/components/display/ApplicantTable.tsx
```diff
@@ -69,7 +81,7 @@ export const ApplicantTable = (props: ApplicantTableProps) => {
               >
                 <td className='pl-6'>{app.applicant_id || 'N/A'}</td>
                 <td className='px-6 py-5'>{app.name}</td>
-                {milestoneText(app.status?.id, app.status?.status)}
+                {milestoneText(app.status?.id, app.status, authUser?.ha_pcn_id)}
```
sunghwan-park-fw, line 84: we don't need to pass app.status.id cause it is included in app.status.

apps/web/src/components/display/ApplicantTable.tsx
```diff
@@ -17,25 +18,36 @@ export const isHired = (id?: number) => {
...
+    return (
+      <td className='px-6 font-bold text-bcGreenHiredText'>
+        Hired
+        <img src={hiredCheckmarkIcon.src} alt='hired checkmark' className='inline-block h-6 mx-2' />
+      </td>
+    );
+  }
+
+  // else check if current user is part of a HA and the status is in the Recruitment step
+  return haId && status?.parent?.id === STATUS.IEN_Recruitment ? (
```
sunghwan-park-fw, line 38: users from hmbc, moh, phsa also can't see the recruitment status. Is it ok?

sunghwan-park-fw (commented):
See my comments

apps/web/src/components/display/ApplicantTable.tsx
coryatwater-FW: hmbc and moh users don't actually get assigned a HA id, as of writing this at least.  So they can see the recruitment status instead of 'In Progress'

apps/web/src/components/display/AddRecordModal.tsx
coryatwater-FW: main reason I added a Field instead of reusing the Select is the longer HA's get cut off in the select, but a full text Field shows them

sunghwan-park-fw (approved):

PR #288, opened 2022-08-05
hide last sync time if user has revoked access

coryatwater-FW:


apps/web/src/components/MenuBar.tsx
```diff
@@ -33,7 +33,7 @@ export const MenuBar: React.FC = () => {
                 </Link>
               ))}
           </div>
-          <LastSyncBar />
+          {!authUser?.revoked_access_date && <LastSyncBar />}
```
sunghwan-park-fw, line 36: How about hiding MenuBar if the user is pending or revoked?

sunghwan-park-fw (approved):

PR #290, opened 2022-08-05
fix incorrect dates for report

coryatwater-FW:


apps/web/src/components/reporting/ReportTable.tsx
```diff
@@ -44,7 +44,8 @@ export const ReportTable = () => {
   };
 
   const download = async (period: Period) => {
-    const from = order === 'ASC' ? periods[0].from : periods[periods.length - 1].from;
```
sunghwan-park-fw, line 47: This should be reverted cause line 83 is enough.

sunghwan-park-fw (changes requested):
See the inline comment.

apps/web/src/components/reporting/ReportTable.tsx
coryatwater-FW: this just causes all the start dates to be Apr 01, 2021

sunghwan-park-fw (approved):

PR #293, opened 2022-08-08
change functionality of hired banners depending on ha of current user

coryatwater-FW:


apps/web/src/components/applicant/ApplicantProfile.tsx
```diff
@@ -1,11 +1,42 @@
...
+  const { applicant, offerWithDiffHa } = useApplicantContext();
+
+  // check for an accepted offer within the same HA as current user
+  const acceptedJobOfferWithSameHa = (): boolean | undefined => {
+    return applicant.jobs?.some(s => s.status_audit?.find(a => isHired(a.status.id)));
+  };
+
+  const showHiredBanner = () => {
+    // show if an offer was accepted within a different HA than logged in user
+    if (offerWithDiffHa) {
```
sunghwan-park-fw, line 19: How about moving this conditional controls to OfferAcceptedBanner component? Then we don't need this helper method.

packages/common/src/helper/is-hired.ts
```diff
@@ -0,0 +1,16 @@
...
+};
+
+export const hasAcceptedOfferWithDiffHa = (
+  jobs: ApplicantJobRO[],
+  currentHa?: number | null,
+): boolean => {
+  // filter out jobs that are from current HA
+  const filteredJobs = jobs.filter(j => j.ha_pcn.id !== currentHa);
+
+  return filteredJobs.filter(j => j.status_audit?.some(s => isHired(s.status.id))).length > 0;
```
sunghwan-park-fw, line 15: Why don't you use `some` instead of `filter`?

sunghwan-park-fw (commented):
See my comments

apps/web/src/components/milestone-logs/recruitment/OfferAcceptedBanner.tsx
```diff
@@ -1,8 +1,37 @@
...
+    // check if applicant accepted offer within same HA as current user
+  } else if (acceptedJobOfferWithSameHa()) {
+    return (
+      <div className={`${defaultStyle} bg-bcGreenHiredText text-white`}>
+        <img src={hiredIndCheckmark.src} alt='add' className='mx-3 h-5' />
+        This applicant has already accepted your job offer and has been hired.&nbsp;
+      </div>
+    );
+  } else {
+    // null if no accepted offers exist
```
sunghwan-park-fw, line 34: It's always a big question to figure out the best place to put a control.
But considering separation of concerns, I like to put the logic to decide to show this component or not, in its parent component, ApplicantProfile.

apps/web/src/components/milestone-logs/recruitment/OfferAcceptedBanner.tsx
```diff
@@ -1,8 +1,37 @@
...
   const defaultStyle = `flex items-center h-14 mb-5 font-bold rounded`;
 
-  return <div className={`${defaultStyle} ${className}`}>{children}</div>;
+  // check for an accepted offer within the same HA as current user
+  const acceptedJobOfferWithSameHa = (): boolean | undefined => {
+    return applicant.jobs?.some(s => s.status_audit?.find(a => isHired(a.status.id)));
+  };
+
+  // show if an offer was accepted within a different HA than logged in user
+  if (offerWithDiffHa) {
```
sunghwan-park-fw, line 17: If possible, we'd better to have one component structure of html elements. It means we return one structure with dynamic properties.
With your approach, it is hard to understand the differences of these branches.

apps/web/src/components/milestone-logs/recruitment/OfferAcceptedBanner.tsx
```diff
@@ -0,0 +1,37 @@
...
+import hiredIndCheckmark from '@assets/img/hired_checkmark_banner.svg';
+import { isHired } from '@ien/common';
+
+export const OfferAcceptedBanner: React.FC = () => {
+  const { applicant, offerWithDiffHa } = useApplicantContext();
+
+  const defaultStyle = `flex items-center h-14 mb-5 font-bold rounded`;
+
+  // check for an accepted offer within the same HA as current user
+  const acceptedJobOfferWithSameHa = (): boolean | undefined => {
```
sunghwan-park-fw, line 12: it seems inconsistent that `offerWithDiffHa` is set in `ApplicantConext` but `offerWithSameHa` is calculated here.
How about letting context expose `hiredHa` and comparing it with user's ha here?
Fundamentally, it is caused by making ApplicantContext depend on AuthContext. I am not sure it is good or bad. Anyway, I like to put those two calculations together.

sunghwan-park-fw (commented):
I left comments as a kind of discussion.

apps/web/src/components/milestone-logs/recruitment/OfferAcceptedBanner.tsx
coryatwater-FW: yup, good point, I had it set up this way when we originally required the full info of the job for a modal, but will redo this chunk and just expose HA instead

sunghwan-park-fw (approved):

PR #304, opened 2022-08-12
add delete job

coryatwater-FW:


apps/api/src/migration/1660225963147-AddDeleteCascadeToJobStatusAudit.ts
```diff
@@ -0,0 +1,23 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "ien_applicant_status_audit" DROP CONSTRAINT "FK_ec2940cc497a06ba811763be464"`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "ien_applicant_status_audit" ADD CONSTRAINT "FK_ec2940cc497a06ba811763be464" FOREIGN KEY ("job_id") REFERENCES "ien_applicant_jobs"("id") ON DELETE CASCADE ON UPDATE NO ACTION`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
```
sunghwan-park-fw, line 15: Is this right that `up` and `down` are same?

apps/web/src/services/applicant.ts
```diff
@@ -108,6 +108,18 @@ export const updateJobRecord = async (
...
 
+export const deleteJobRecord = async (
+  user_id?: string | null,
+  status_id?: string,
+): Promise<void> => {
+  if (!user_id) return;
+  try {
+    await axios.delete(`/ien/${user_id}/job/${status_id}`);
+  } catch (e) {
+    notifyError(e as AxiosError);
```
sunghwan-park-fw, line 119: You don't need to call notifyError here. All axios errors are notified by axios interceptor

apps/web/src/components/milestone-logs/recruitment/Record.tsx
```diff
@@ -98,6 +106,33 @@ export const Record: React.FC<RecordProps> = ({
...
+  const handleDeleteJob = (jobId?: string) => {
+    if (jobId) {
+      deleteJob(jobId);
+    }
+
+    setDeleteModalVisible(false);
+  };
+
+  const deleteButton = () => {
+    return canDelete(authUser?.user_id, added_by?.id) ? (
```
sunghwan-park-fw, line 118: It might a rare case but what if users with applicant write access added a job and then they lose the access later? can they still see the delete button?

sunghwan-park-fw (approved):
See my comments

apps/api/src/migration/1660225963147-AddDeleteCascadeToJobStatusAudit.ts
coryatwater-FW: up contains 'ON DELETE CASCADE', down contains 'ON DELETE NO ACTION'

apps/web/src/components/milestone-logs/recruitment/Record.tsx
coryatwater-FW: no, it'd be blocked by the role mask

PR #306, opened 2022-08-12
IEN 537 ats applicant sync

dbayly-freshworks:
Added a route to return all applicants (and related tables) 
Also added a JWT guard.

### DIFF
--- a/apps/api/src/applicant/external-api.service.ts
+++ b/apps/api/src/applicant/external-api.service.ts
@@ -629,7 +629,7 @@ export class ExternalAPIService {
   }
 
   async getApplicants() {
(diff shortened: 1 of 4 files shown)


sunghwan-park-fw (approved):

apps/api/src/applicant/external-api.service.ts
```diff
@@ -627,4 +627,17 @@ export class ExternalAPIService {
       return this.ienMasterService.ienUsersRepository.findAndCount(query);
     }
   }
+
+  async getApplicants() {
+    return await this.ienapplicantRepository.find({
```
sunghwan-park-fw, line 632: Sonar complains of this return with `await`.

PR #321, opened 2022-08-18
IEN-626 | hide filter tab on manage applicants page

sunghwan-park-fw:


apps/web/src/components/user/AclMask.tsx
```diff
@@ -1,9 +1,11 @@
 import { PropsWithChildren, ReactNode } from 'react';
 import { Access, hasAccess } from '@ien/common';
 import { useAuthContext } from '../AuthContexts';
+import * as React from 'react';
```
coryatwater-FW, line 4: do we need this line?

apps/web/src/utils/is-ha-users.ts
```diff
@@ -0,0 +1,9 @@
+import { Authorities, EmployeeRO, RoleSlug } from '@ien/common';
+
+export const isHaUsers = (user?: EmployeeRO): boolean => {
+  return (
+    !!user &&
+    user.roles.every(role => role.slug !== RoleSlug.Admin) &&
+    user.organization !== Authorities.HMBC.name
```
coryatwater-FW, line 7: what if the user is MoH?

coryatwater-FW (approved):
couple comments, but looks good to me

apps/web/src/utils/is-ha-users.ts
sunghwan-park-fw: MoH has only reporting role by default so then don't have higher privileges than HA.

apps/web/src/components/user/AclMask.tsx
sunghwan-park-fw: It's why we shouldn't do `copy and paste`

PR #325, opened 2022-08-19
IEN - 537 continued 

dbayly-freshworks:
Added expiry to tokens and added date filter to records

### DIFF
--- a/apps/api/src/applicant/external-api.service.ts
+++ b/apps/api/src/applicant/external-api.service.ts
@@ -22,7 +22,6 @@ import { SyncApplicantsAudit } from './entity/sync-applicants-audit.entity';
 import { IENMasterService } from './ien-master.service';
 import { IENUsers } from './entity/ienusers.entity';
 import { IENUserFilterAPIDTO } from './dto/ienuser-filter.dto';
-import dayjs from 'dayjs';
 
 @Injectable()
 export class ExternalAPIService {
--- a/apps/api/src/applicant/external-api.service.ts
+++ b/apps/api/src/applicant/external-api.service.ts
@@ -9,6 +9,7 @@ import {
   FindManyOptions,
   ObjectLiteral,
   SelectQueryBuilder,
+  Between,
 } from 'typeorm';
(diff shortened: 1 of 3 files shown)


apps/api/src/applicant/external-api.service.ts
```diff
@@ -628,16 +631,26 @@ export class ExternalAPIService {
...
-        'applicant_audit',
-      ],
-    });
+
+  
+  async getApplicants(start?:string,end?:string ) {
+      return this.ienapplicantRepository.find({
+        where: {
+          updated_date: Between(
+            new Date(start ||'1918-07-18' ).toISOString(),
```
sunghwan-park-fw, line 640: Does this date have special meaning?

dbayly-freshworks: It's the day World war 1 was declared. I needed an arbitrary date that should be well before any record is updated. 
It's an easter egg, and I can remove it if you like.

apps/api/src/auth/jwt.guard.ts
```diff
@@ -41,11 +42,20 @@ export class JWTGuard implements CanActivate {
...
         process.env.JWT_SECRET ? process.env.JWT_SECRET : 'jwtsecret',
-      );
-      return decoded;
+      )) as any;
+      if(!decoded.exp){
+        return false; 
+      }
+      // Check token expiry
+      // JWT Token expiry is in seconds since the unix epoc, so we need to multiply them by 1000 to convert them into timestamps.
+      if(dayjs(decoded?.exp*1000).isAfter(dayjs()) && dayjs(decoded?.exp*1000).isBefore(dayjs().add(5,'minutes')) ){
```
sunghwan-park-fw, line 54: Is this to enforce token to expire in 5 minutes?

dbayly-freshworks: That is correct. I check if the expiry is after the current time, but before five minutes in the future.

sunghwan-park-fw (approved):

PR #333, opened 2022-08-24
fix selected value for date picker

coryatwater-FW:
JS Date object with yyyy-MM-dd formatting causes it to be reduced by 1 day.  dayjs and toDate corrects this


dbayly-freshworks (approved):
looks fine to me!

PR #334, opened 2022-08-25
show latest milestone for hmbc users

coryatwater-FW:
IEN-604 - fix hired banner on Details page and hired row on Manage Applicants page showing 'when candidate was not selected' option was added

IEN-636 - when HMBC user is logged in, show applicants latest milestone in process regardless of having accepted a job

* HA's will see Hired regardless of latest milestone of applicant, HMBC will see latest milestone


dbayly-freshworks (approved):
Looks ok to me

PR #338, opened 2022-08-26
Ien 537 ats sync final updated (hopefully)

dbayly-freshworks:
Split the query into steps. 
The first step fetches ids of applicants that have milestones updated within the to/from date fields. 
The second fetches all the applicants and related tables with those ids. 

Also updated the query params to use `IENUserFilterAPIDTO ` so to add pagination functionality to the query.

### DIFF
--- a/apps/api/src/applicant/external-api.service.ts
+++ b/apps/api/src/applicant/external-api.service.ts
@@ -657,11 +657,11 @@ export class ExternalAPIService {
     WHERE updated_date <'${to || dayjs().add(1, 'day').format('YYYY-MM-DD')}' and updated_date >'${
       from || '1914-07-18'
     }'`;
-    if(limit){
-      queryString = queryString +(limit? ` LIMIT ${limit} `:'' )
+    if (limit) {
+      queryString = queryString + (limit ? ` LIMIT ${limit} ` : '');
     }
-    if(skip){
-      queryString = queryString + (skip?` OFFSET ${skip} `:'');
+    if (skip) {
+      queryString = queryString + (skip ? ` OFFSET ${skip} ` : '');
     }
     const ids: string[] = (await this.ienapplicantStatusAuditRepository.query(queryString)).map(
       (result: { applicant_id: string }) => result.applicant_id,
--- a/apps/api/src/applicant/external-api.service.ts
+++ b/apps/api/src/applicant/external-api.service.ts
@@ -658,10 +658,10 @@ export class ExternalAPIService {
       from || '1914-07-18'
     }'`;
     if (limit) {
-      queryString = queryString + (limit ? ` LIMIT ${limit} ` : '');
+      queryString = queryString +` LIMIT ${limit} `;
     }
     if (skip) {
-      queryString = queryString + (skip ? ` OFFSET ${skip} ` : '');
+      queryString = queryString + `OFFSET ${skip} `;
     }
     const ids: string[] = (await this.ienapplicantStatusAuditRepository.query(queryString)).map(
       (result: { applicant_id: string }) => result.applicant_id,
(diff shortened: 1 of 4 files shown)


apps/api/src/applicant/external-api.service.ts
```diff
@@ -650,14 +649,25 @@ export class ExternalAPIService {
...
+      return [];
+    }
     return this.ienapplicantRepository.find({
-      where: {
-        updated_date: Between(
-          new Date(start || '1918-07-18').toISOString(),
-          end ? new Date(end).toISOString() : new Date().toISOString(),
-        ),
+      where: (qb: any) => {
+        qb.where(`IENApplicant.id IN (:...ids)`, { ids });
```
sunghwan-park-fw, line 667: How about using the sub list of ids instead of `skip` and `take`?

dbayly-freshworks: Thats a great idea, it should speed up the query.

apps/api/src/applicant/external-api.service.ts
```diff
@@ -650,13 +649,28 @@ export class ExternalAPIService {
...
-  async getApplicants(start?: string, end?: string) {
+  async getApplicants(filter: IENUserFilterAPIDTO) {
+    const { from, to, limit, skip } = filter;
+    let queryString = `SELECT DISTINCT applicant_id  
+    FROM ien_applicant_status_audit 
+    WHERE updated_date <'${to || dayjs().add(1, 'day').format('YYYY-MM-DD')}' and updated_date >'${
+      from || '1914-07-18'
+    }'`;
+    if (limit) {
+      queryString = queryString + (limit ? ` LIMIT ${limit} ` : '');
```
sunghwan-park-fw, line 660: with `if (limit)`, `limit ?` seems redundant.

dbayly-freshworks: You're right 😂 . When I wrote the line first, I had it as part of the declaration for QueryString, so I had to make sure it was null safe. I'll fix this.

apps/api/src/applicant/external-api.service.ts
```diff
@@ -651,13 +650,28 @@ export class ExternalAPIService {
     }
   }
 
-  async getApplicants(start?: string, end?: string) {
+  async getApplicants(filter: IENUserFilterAPIDTO) {
+    const { from, to, limit, skip } = filter;
+    let queryString = `SELECT DISTINCT applicant_id
```
arranfw, line 655: Why are we building a string like this rather than using queryBuilder? I'm fairly certain this could open the query up to SQL injections

arranfw (changes requested):

apps/api/src/applicant/external-api.service.ts
dbayly-freshworks: We do have some protections in place to protect against that (validator and the auth guard), but I am happy to rewrite this to use the query builder to be safe.

apps/api/src/applicant/external-api.service.ts
```diff
@@ -651,22 +650,34 @@ export class ExternalAPIService {
...
+        .select('applicant_id')
+        .where('updated_date < :toDate AND updated_date > :fromDate', {
+          toDate: to || dayjs().add(1, 'day').format('YYYY-MM-DD'),
+          fromDate: from || '1914-07-18',
+        })
+        .limit(limit)
+        .skip(skip)
+        .distinct()
+        .execute()
+    ).map((result: { applicant_id: string }) => {
```
sunghwan-park-fw, line 668: How about adding `where` clause with `NotNull` instead of this `map`?

sunghwan-park-fw (approved):

arranfw (approved):

PR #339, opened 2022-08-26
IEN-643, IEN-644 fix hired status of job

sunghwan-park-fw:


dbayly-freshworks (approved):
Looks ok to me. Always makes me happy to see a pr with more deletions than additions

PR #349, opened 2022-09-07
fix manage applicant page tabs not returning data

coryatwater-FW:
- also includes fix for landing tab for Applicant Details page

### DIFF
--- a/apps/api/src/applicant/ienapplicant.util.service.ts
+++ b/apps/api/src/applicant/ienapplicant.util.service.ts
@@ -145,20 +145,19 @@ export class IENApplicantUtilService {
     const status_list = status.split(',');
     const parent_status = await this.ienMasterService.ienApplicantStatusRepository.find({
       where: {
-        id: In(status_list),
+        category: In(status_list),
         parent: IsNull(),
       },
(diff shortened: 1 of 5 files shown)


apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -145,20 +145,19 @@ export class IENApplicantUtilService {
     const status_list = status.split(',');
     const parent_status = await this.ienMasterService.ienApplicantStatusRepository.find({
       where: {
-        id: In(status_list),
+        category: In(status_list),
```
sunghwan-park-fw, line 148: we don't need to check parent any more.

apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -145,20 +145,19 @@ export class IENApplicantUtilService {
...
         parent: IsNull(),
       },
-      relations: ['children'],
     });
+
+    const status_to_return: string[] = [];
+
```
sunghwan-park-fw, line 154: can we `return parent_status.map(({id}) => id)`

PR #350, opened 2022-09-07
WIP - Updated seeding script

dbayly-freshworks:


sunghwan-park-fw (commented):
this might conflict with IEN-648_fix-migrations

PR #359, opened 2022-09-12
update report queries with new milestone ids

coryatwater-FW:
- added enum for status Id's
- makes queries more readable
- centralizes data location for easy changes


apps/api/src/report/report.util.service.ts
```diff
@@ -229,22 +231,54 @@ export class ReportUtilService {
...
-      --col03 - 'Completed English Language Requirement' status: 8
-      --col04 - 'Referred to NCAS' status: 9,10,11
-      --col05 - 'Completed NCAS (BCCNM Assessment)' status: 12
-      --col06 - 'Completed Additional Education' status: 13, 14
-      --col07 - 'Referred to NCLEX' status: 15,17
-      --col08 - 'NCLEX - Passed' status: 18
-      --col09 - 'BCCNM Licensed - Full License' status: 19
-      --col10 - 'Granted full licensure' status: 19
-      --col11 - 'Granted provisional licensure' status: 16
+      --col01 - 'Applied to NNAS' status: 4, 5, 6 e68f902c-8440-4a9c-a05f-2765301de800
```
sunghwan-park-fw, line 234: Even though we provide these comments for better understanding, it is hard to grasp the meaning of IDs in the query.
How about passing a map with keys as status and values as id to each report util function?

apps/api/src/report/report.util.service.ts
```diff
@@ -229,22 +231,54 @@ export class ReportUtilService {
...
+      --col27 - 'Applicant Referred to VIHA' 001dfb24-1618-e975-6578-3a0601aac804
+
+      --col29 - 'Granted full licensure' status: 19 632374e6-ca2f-0baa-f994-3a05e77c118a, 18aa32c3-a6a4-4431-8283-89931c141fde
+      --col30 - 'Granted provisional licensure' status: 16 91f55faa-c71d-83c8-4f10-3a05e778afbc, d2656957-ec58-15c9-1e21-3a05e778dc8e
       WITH active_applicants AS (
         SELECT
           t1.*,
-          COALESCE((SELECT max(sa.status_id) FROM public.ien_applicant_status_audit sa 
-          WHERE sa.applicant_id=t1.id AND sa.status_id IN (19, 16)), 0) AS has_license
+          COALESCE((SELECT sa.status_id FROM public.ien_applicant_status_audit sa
```
sunghwan-park-fw, line 273: this COALESCE block is confusing. How about adjusting indents?

sunghwan-park-fw (commented):
I left comments.

packages/common/src/enum/milestone-status.ts
```diff
@@ -65,3 +65,60 @@ export const COMPLETED_STATUSES = [
   STATUS.JOB_COMPETITION_CANCELLED,
   STATUS.JOB_OFFER_ACCEPTED,
 ];
+
+export enum StatusId {
+  COMPLETED_SLA = '9066b792-6fbf-4e60-803f-0554e4b4dba9',
```
sunghwan-park-fw, line 70: What I was thinking was to generate a map object from a database query.
The reason I was wary of your idea of changing status enum is that it is hard to notice whether there is an error or not.
Considering its complexity, uuid must never be meant to be written manually.
We may refactor this later.

sunghwan-park-fw (approved):

PR #383, opened 2022-09-19
add transaction to queries

coryatwater-FW:


apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -151,15 +152,20 @@ export class IENApplicantUtilService {
     applicant: IENApplicant,
     dataToUpdate: any,
     job: IENApplicantJob | null,
-  ): Promise<IENApplicantStatusAudit> {
+  ): Promise<IENApplicantStatusAudit | any> {
```
dbayly-freshworks, line 155: Out of curiosity, why the `any` here

dbayly-freshworks (approved):
One nit, otherwise looks ok to me

apps/api/src/applicant/ienapplicant.service.ts
```diff
@@ -334,10 +336,12 @@ export class IENApplicantService {
...
     }
-    await this.ienapplicantStatusAuditRepository.save(status_audit);
 
-    await this.ienapplicantRepository.update(status_audit.applicant.id, {
-      updated_date: new Date(),
+    await getManager().transaction(async manager => {
+      await manager.save<IENApplicantStatusAudit>(status_audit);
+      await manager.update<IENApplicant>(IENApplicant, status_audit.applicant.id, {
+        updated_date: new Date(),
+      });
```
sunghwan-park-fw, line 344: shouldn't we include `updateLatestStatusOnApplicant` below?

apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -151,15 +152,20 @@ export class IENApplicantUtilService {
...
     const status: Partial<IENApplicantStatusAudit> = {
       applicant: applicant,
       job: job,
       ...dataToUpdate,
     };
-    const status_audit = this.ienapplicantStatusAuditRepository.create(status);
-    return this.ienapplicantStatusAuditRepository.save(status_audit);
+
+    await getManager().transaction(async manager => {
+      const status_audit = manager.create<IENApplicantStatusAudit>(IENApplicantStatusAudit, status);
```
sunghwan-park-fw, line 164: `create` just creates an entity to be saved, so these two lines are just one operation.

sunghwan-park-fw (commented):
I believe there are a lot more points requiring transaction.

apps/api/src/applicant/ienapplicant.util.service.ts
coryatwater-FW: was mainly just to test, the return is inside the transaction now, so needed to return something, can change it

apps/api/src/applicant/external-api.service.ts
```diff
@@ -294,13 +302,16 @@ export class ExternalAPIService {
...
     additional_data: object | undefined | unknown = undefined,
+    manager?: EntityManager,
   ): Promise<SyncApplicantsAudit> {
     if (id) {
       const audit = await this.syncApplicantsAuditRepository.findOne(id);
       if (audit) {
         audit.is_success = is_success;
         audit.additional_data = additional_data;
-        await this.syncApplicantsAuditRepository.save(audit);
+        manager
```
sunghwan-park-fw, line 312: we can always pass manager.

coryatwater-FW: saveSyncApplicantsAudit gets called in the catch block of the try/catch, outside the transaction, I suppose we could re-organize it or just remove that catch call?

sunghwan-park-fw: You're right. we can use `finally` block to log the result instead of `catch`.

apps/api/src/applicant/ienapplicant.util.service.ts
```diff
@@ -270,8 +279,12 @@ export class IENApplicantUtilService {
     return this.ienMasterService.ienJobLocationRepository.findByIds(ids);
   }
 
-  async updateLatestStatusOnApplicant(mappedApplicantList: string[]): Promise<void> {
+  async updateLatestStatusOnApplicant(
+    mappedApplicantList: string[],
+    manager?: EntityManager,
+  ): Promise<void> {
     try {
+      const entityManager = manager ? manager : getManager();
```
sunghwan-park-fw, line 287: how about `manager || getManager()`?

apps/api/src/applicant/ienapplicant.service.ts
```diff
@@ -231,14 +240,16 @@ export class IENApplicantService {
 
     data.status = status_obj;
 
-    let job = null;
+    // needed specfic typing as job is now being used inside nested transaction
+    let job = new IENApplicantJob();
+
```
sunghwan-park-fw, line 245: how about this? ienapplicantUtilService.getJob should handle the case that job_id is not defined or throw an exception.
```
const job = await this.ienapplicantUtilService.getJob(job_id);
if (id !== job?.applicant.id) {
  throw new BadRequestException('Provided applicant and competition/job does not match');
}
```

coryatwater-FW: fixed

sunghwan-park-fw (approved):

PR #386, opened 2022-09-19
Ien 691 fixing reports

dbayly-freshworks:
Updated nursing educations column to have a default value. 
This should fix the issues with the education report


apps/api/src/applicant/entity/ienapplicant.entity.ts
```diff
@@ -52,7 +52,7 @@ export class IENApplicant {
   @Column('varchar', { nullable: true })
   pr_status?: string;
 
-  @Column('jsonb', { nullable: true })
+  @Column('jsonb', { nullable: false, default: `[]` })
```
sunghwan-park-fw, line 55: In which case does it cause a problem? considering nursing_educations is optional, I think null shouldn't be an issue.

dbayly-freshworks:
Closing this PR because it was only effecting my local env, and because this PR was causing more problems than it solved.

PR #387, opened 2022-09-19
IEN-691 fix report 4 when an applicant has multiple license milestones

sunghwan-park-fw:


bhumin-fw (approved):
This looks good. Please go ahead and merge.
another approach if you have more than 5 such data to capture. (in this query no requirement to follow the below approach it is just for your reference)
first, take all data where status is in (list of status) and run the group by status. Here, it will fetch status-wise count in one query run (in place of a separate query run for each status)

All data comes in a row, but your requirement is to put them in columns. to do so you have to apply one more query here that converts row data in the column and adds the missing value with 0.

coryatwater-FW (approved):

PR #390, opened 2022-09-20
IEN-695 delete duplicate milestones

sunghwan-park-fw:


dbayly-freshworks (approved):
~Nit: Might be good to populate the down method. This change is reversable

sunghwan-park-fw:
> ~Nit: Might be good to populate the down method. This change is reversable

I spent some time while thinking of what the down function should be like and I wasn't sure if I need to restore the duplicate statuses without restoring related milestones of applicants. Down function seems like one of the most difficult topics of typeorm.

PR #400, opened 2022-09-23
IEN 707 - reorder max date check for data extract

coryatwater-FW:
re ordered the max date check for the datapicker on Data Extract, was previously giving the current value picked priority

### DIFF
--- a/apps/web/src/components/reporting/DataExtractReport.tsx
+++ b/apps/web/src/components/reporting/DataExtractReport.tsx
@@ -81,7 +81,7 @@ export const DataExtractReport = () => {
                   label='End Date'
                   format='yyyy-MM-dd'
                   min={dayjs(values.from || MIN_DATE).toDate()}
-                  max={dayjs(values.to || getMaxDate()).toDate()}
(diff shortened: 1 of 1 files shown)


apps/web/src/components/reporting/DataExtractReport.tsx
```diff
@@ -81,7 +81,7 @@ export const DataExtractReport = () => {
                   label='End Date'
                   format='yyyy-MM-dd'
                   min={dayjs(values.from || MIN_DATE).toDate()}
-                  max={dayjs(values.to || getMaxDate()).toDate()}
+                  max={dayjs(getMaxDate() || values.to).toDate()}
```
sunghwan-park-fw, line 84: I don't remember why I needed `values.to` condition.

coryatwater-FW: I'll try removing it and see what happens

sunghwan-park-fw: getMaxDate always returns a value, so we don't need it.

sunghwan-park-fw (approved):

PR #401, opened 2022-09-23
IEN 708 - fix licenses count in report 4

coryatwater-FW:
I figured since we're already getting these counts at the beginning of the query and reusing them for the total count, to just to reuse them in the query again for each

### DIFF
--- a/apps/api/src/report/report.util.service.ts
+++ b/apps/api/src/report/report.util.service.ts
@@ -313,10 +313,10 @@ export class ReportUtilService {
       SELECT 'NCLEX - Passed' as status, count(*) FROM report WHERE status_id = '${StatusId.NCLEX_PASSED}' UNION ALL
       SELECT 'REx-PN – Written' as status, count(*) FROM report WHERE status_id = '${StatusId.REX_PN_WRITTEN}' UNION ALL
       SELECT 'REx-PN – Passed' as status, count(*) FROM report WHERE status_id = '${StatusId.REX_PN_PASSED}' UNION ALL
-      SELECT 'BCCNM Full Licence LPN' as status, count(*) FROM report WHERE status_id = '${StatusId.BCCNM_FULL_LICENCE_LPN}' UNION ALL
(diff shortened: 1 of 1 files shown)


apps/api/src/report/report.util.service.ts
```diff
@@ -313,10 +313,10 @@ export class ReportUtilService {
       SELECT 'NCLEX - Passed' as status, count(*) FROM report WHERE status_id = '${StatusId.NCLEX_PASSED}' UNION ALL
       SELECT 'REx-PN – Written' as status, count(*) FROM report WHERE status_id = '${StatusId.REX_PN_WRITTEN}' UNION ALL
       SELECT 'REx-PN – Passed' as status, count(*) FROM report WHERE status_id = '${StatusId.REX_PN_PASSED}' UNION ALL
-      SELECT 'BCCNM Full Licence LPN' as status, count(*) FROM report WHERE status_id = '${StatusId.BCCNM_FULL_LICENCE_LPN}' UNION ALL
-      SELECT 'BCCNM Full Licence RN' as status, count(*) FROM report WHERE status_id = '${StatusId.BCCNM_FULL_LICENSE_RN}' UNION ALL
-      SELECT 'BCCNM Provisional Licence LPN' as status, count(*) FROM report WHERE status_id = '${StatusId.BCCNM_PROVISIONAL_LICENSE_LPN}' UNION ALL
-      SELECT 'BCCNM Provisional Licence RN' as status, count(*) FROM report WHERE status_id = '${StatusId.BCCNM_PROVISIONAL_LICENSE_RN}' UNION ALL
+      SELECT 'BCCNM Full Licence LPN' as status, count(*) FROM report WHERE full_lpn > 0 UNION ALL
```
sunghwan-park-fw, line 316: It's reasonable, but I don't understand how it could fix the problem.

coryatwater-FW: I don't fully understand the issue, but the total columns for each full and provisional licensure has the correct amount, while the singular columns for each don't.  So I just took the value that was correct

sunghwan-park-fw: Oh, it's because that status_id means the latest milestone in licensing/registration stage. If an applicant with a license has other milestone later than it, it won't be counted.

sunghwan-park-fw (approved):

PR #404, opened 2022-09-23
IEN-710 fix report 5 - exclude inactive(withdrew, hired) applicant

sunghwan-park-fw:


dbayly-freshworks (approved):
It took a bit to get through, but this looks good.

PR #407, opened 2022-09-23
Updated sync back to ATS's response object to new format

dbayly-freshworks:


apps/api/src/applicant/external-api.service.ts
```diff
@@ -666,15 +667,32 @@ export class ExternalAPIService {
...
         'applicant_status_audit.status',
         'applicant_status_audit.job',
+        'applicant_status_audit.updated_by',
+        'applicant_status_audit.added_by',
       ],
     });
+
+    return results.map((result): ApplicantSyncRO => {
+      let updated_date = result.updated_date;
+      result.applicant_status_audit.map(status => {
```
sunghwan-park-fw, line 685: We update `applicant.updated_date` when updating a milestone but not adding a milestone.
If we do when adding too, don't we need this routine to calculate the latest updated_date of milestones?

sunghwan-park-fw (approved):

PR #413, opened 2022-09-27
fix no data in nnas row, remove redundant status checks

coryatwater-FW:


apps/api/src/report/report.util.service.ts
```diff
@@ -835,7 +832,6 @@ export class ReportUtilService {
               AND asa.start_date >= sdoes.nnas
```
coryatwater-FW, line 832: @sunghwan-park-fw  @dbayly-freshworks  I wonder, should this be comparing to sdoes.bccnm_ncas isntead of sdoes.nnas?

sunghwan-park-fw: No, we try to find the start of the next stage, bccnm_ncas, so it should be later than nnas.

sunghwan-park-fw (approved):

PR #417, opened 2022-09-28
add withdrew from competition check to report 9

coryatwater-FW:
Added Withdrew from Competition Outcome check to Report 9

### DIFF
--- a/apps/api/src/report/report.util.service.ts
+++ b/apps/api/src/report/report.util.service.ts
@@ -760,8 +760,7 @@ export class ReportUtilService {
         FROM public.ien_applicant_status_audit iasa
         LEFT JOIN public.ien_applicant_jobs iaj ON iasa.job_id=iaj.id
         WHERE iasa.status_id 
-        IN ('${StatusId.WITHDREW_FROM_PROGRAM}', '${StatusId.JOB_OFFER_ACCEPTED}')
-        OR iasa.status_id = '${StatusId.WITHDREW_FROM_COMPETITION}'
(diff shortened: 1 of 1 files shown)


apps/api/src/report/report.util.service.ts
```diff
@@ -759,28 +759,31 @@ export class ReportUtilService {
               ROW_NUMBER() OVER(PARTITION BY iasa.applicant_id ORDER BY iasa.start_date DESC, iasa.updated_date DESC) AS rank
         FROM public.ien_applicant_status_audit iasa
         LEFT JOIN public.ien_applicant_jobs iaj ON iasa.job_id=iaj.id
-        WHERE iasa.status_id IN ('${StatusId.WITHDREW_FROM_PROGRAM}', '${StatusId.JOB_OFFER_ACCEPTED}') AND iasa.start_date::date <= '${to}'
+        WHERE iasa.status_id 
+        IN ('${StatusId.WITHDREW_FROM_PROGRAM}', '${StatusId.JOB_OFFER_ACCEPTED}', '${StatusId.WITHDREW_FROM_COMPETITION}')
```
sunghwan-park-fw, line 763: Do we need to check if the reason is `withdrew from ien program`?

coryatwater-FW: Docs says we only calculate 'Hired' applicants and exclude any active or withdrawn applicants, so I generalized it.  Can confirm though

sunghwan-park-fw: Even though an applicant withdraws from a competition, it could be active in other competitions.

coryatwater-FW: this report ignores active applicants so they won't be included in the data anyways

coryatwater-FW:
closing, not needed anymore

PR #418, opened 2022-09-28
Revert Sync timer back to being daily

dbayly-freshworks:


sunghwan-park-fw (approved):
Don't we need to uncomment `make apply` in github actions?

sunghwan-park-fw (approved):

PR #427, opened 2022-10-03
IEN-681 remove status_id enum, add an endpoint to retrieve all report…

sunghwan-park-fw:


coryatwater-FW (approved):
looks good to me

apps/api/src/report/report.controller.ts
```diff
@@ -5,6 +5,9 @@ import { AuthGuard } from 'src/auth/auth.guard';
 import { ReportService } from './report.service';
 import { AppLogger } from 'src/common/logger.service';
 import { AllowAccess } from 'src/common/decorators';
+import { InjectRepository } from '@nestjs/typeorm';
+import { IENApplicantStatus } from '../applicant/entity/ienapplicant-status.entity';
+import { Repository } from 'typeorm';
```
coryatwater-FW, line 10: I know it's not present in many files(should organize at some point), but we should organize the imports consistently.  I tend to do 3rd party at the top, internal below, but also open to other ways

sunghwan-park-fw: I'll try this option.
https://blog.jetbrains.com/webstorm/2018/05/optimize-imports-in-webstorm/

PR #428, opened 2022-10-03
IEN 735 - add e2e test for report 1

coryatwater-FW:


sunghwan-park-fw (commented):
How about defining a spec file for each report?

apps/api/test/reports.e2e-spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    app = moduleFixture.createNestApplication();
+    await app.init();
+  });
+
+  afterAll(async () => {
+    await app.close();
+  });
+
+  it('Add Applicant /ien (POST)', done => {
```
sunghwan-park-fw, line 36: `it` defines a test case. so, these two cases should be moved into the last one.

apps/api/test/applicant.e2e-spec.ts
```diff
@@ -91,7 +91,7 @@ describe('ApplicantController (e2e)', () => {
       .get('/ien')
       .expect(res => {
         const { body } = res;
-        expect(body[1]).toBe(2);
+        expect([2, 4]).toContain(body[1]);
```
sunghwan-park-fw, line 94: I wonder why body[1] could be 2 or 4

coryatwater-FW: this is something I wanted to bring up in SU and discuss with you

sunghwan-park-fw (approved):

PR #429, opened 2022-10-03
IEN-745 0 day duration

sunghwan-park-fw:


apps/web/src/components/milestone-logs/MilestoneTable.tsx
```diff
@@ -103,7 +103,7 @@ export const MilestoneTable = ({ category }: MilestoneTableProps) => {
       }
     }
 
-    if (start === end) return '-';
+    if (start === end) return '0 day';
```
coryatwater-FW, line 106: Think the correct way would be '0 days'

sunghwan-park-fw: It's shame that there is an error in the change of just five characters.

coryatwater-FW: happens to the best of us! :P

coryatwater-FW (approved):

PR #435, opened 2022-10-05
IEN 737 - add test for report 3

coryatwater-FW:


dbayly-freshworks:
It looks like because no application related files were edited, the test suite was not run. 

It might be good to look into updating our config to run the tests regardless, or to make a harmless change to an application file to get them to trigger.

PR #438, opened 2022-10-05
IEN-760 fix milestone duration in milestone table

sunghwan-park-fw:


dbayly-freshworks (approved):
Looks ok, though the map/reduce/sort chain in getNextMilestone is a little hard to parse.

PR #441, opened 2022-10-06
update readme doc

coryatwater-FW:


apps/web/README.md
```diff
@@ -1,34 +1,44 @@
...
+> ** Authentication **\
+> See [Authentication](../../README.md#authentication)
 
-```bash
-npm run dev
-# or
-yarn dev
-```
+## Pages 
+***Each page requires particular roles and each role grants certain privileges, depending on roles, pages may look different***
```
sunghwan-park-fw, line 9: This line seems having two sentences. how about using conjunction or split them?

PR #443, opened 2022-10-06
IEN-759 add details row to report 8

sunghwan-park-fw:


apps/web/src/services/report.ts
```diff
@@ -90,7 +90,7 @@ const createSheet = (
   creator: ReportCreator,
   filter: PeriodFilter,
 ): WorkSheet => {
```
coryatwater-FW, line 92: regardless of the code smell, I feel like this function is getting quite complex

dbayly-freshworks: ☝️ Agreed

apps/web/src/services/report.ts
```diff
@@ -219,9 +223,23 @@ const reportCreators: ReportCreator[] = [
...
+        '',
+        ...Object.keys(data[0])
+          .slice(1)
+          .map(v => {
+            return v.includes('current_period') ? `${v.split(' ')[0]}*` : v;
+          }),
+      ];
+    },
+    colWidths: [35, 20, 20, 20],
+    notes: (data: Record<string, string | number>[]): any[] => {
```
dbayly-freshworks, line 236: Notes is a bit vague for a variable name. It would be good to add a small dock block to explain the intent behind it.

apps/web/src/services/report.ts
```diff
@@ -42,13 +43,12 @@ export const getReportByEOI = async (filter?: PeriodFilter) => {
...
-      ? dayjs(period.from).format('MMM DD')
-      : dayjs(period.from).format('MMM DD, YYYY');
-
-  return `${from} - ${dayjs(period.to).format('MMM DD, YYYY')}`;
+export const getTimeRange = ({ from, to }: PeriodFilter): string => {
+  const start = dayjs(from).format('MMM DD, YYYY');
+  if (to) {
+    return `${start} ~ ${dayjs(to).format('MMM DD, YYYY')}`;
+  }
+  return `${start} ~ `;
```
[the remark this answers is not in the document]
sunghwan-park-fw: It is to mean that the current period is on going. But do you think it is better to show `to` as today?

apps/web/src/services/report.ts
sunghwan-park-fw: I'll add some comments for clarification

apps/web/src/services/report.ts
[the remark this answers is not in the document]
coryatwater-FW: I think something is better than nothing, maybe tack on On Going to better identify it.  Also, question, isn't the Current periods `to` date always set to the current date if it's on going? or is that just for display purposes on the UI

ex. Period started Oct 4th and it's the 6th so the UI shows Oct 4th - Oct 6th

apps/web/src/services/report.ts
```diff
@@ -42,23 +44,25 @@ export const getReportByEOI = async (filter?: PeriodFilter) => {
...
 };
 
-const applyNumberFormat = (sheet: WorkSheet, s: CellAddress, e: CellAddress): void => {
-  for (let r = s.r; r <= e.r; ++r) {
-    for (let c = s.c; c <= e.c; ++c) {
+const applyNumberFormat = (sheet: WorkSheet, rows: any[][]): void => {
```
dbayly-freshworks, line 55: Nit - This function is hard to read, it might be good to add a doc block with a short explaination of it's purpose. 
Also, it would be good to use longer variable names, eg row_number instead of r, and column_number instead of c.

dbayly-freshworks (approved):
This looks ok/functional, but is really hard to read.

apps/web/src/services/report.ts
sunghwan-park-fw: I updated the function and fixed the report 1 test case.

PR #450, opened 2022-10-13
IEN-474 | add report 10

sunghwan-park-fw:


dbayly-freshworks (approved):
This looks ok. My main concern is that an applicant requires a job offer accepted status to be counted for this. I think that that will exclude a large portion of the data.

PR #459, opened 2023-01-31
IEN-779 - update existing report docs

coryatwater-FW:
Added a few updates to the existing reports doc to match SQL queries and report data


dbayly-freshworks (approved):
Looks good, much easier to read now.

PR #461, opened 2023-02-03
IEN-740 Add report 6 test

coryatwater-FW:
Add test cases for Report 6


apps/api/test/06-report-6.e2e-spec.ts
```diff
@@ -0,0 +1,141 @@
...
+      job_id: jobTempId || 'WillHaveJobId',
+      start_date: start_date,
+    });
+
+    const { body } = await request(app.getHttpServer()).post(addStatusUrl).send(milestone);
+    applicantId = id;
+    applicantStatusId = body.id;
+  };
+
+  it('Add Recruitment related statuses for 4 HAs', async () => {
```
sunghwan-park-fw, line 79: is there any reason for the test to be run for 4 HAs?

apps/api/test/06-report-6.e2e-spec.ts
```diff
@@ -0,0 +1,141 @@
...
+      applicant.registration_date = '2022-06-01';
+      const { id } = await addApplicant(applicant);
+
+      await addJob(id, HEALTH_AUTHORITIES[i].id, i.toString());
+      await addMilestone(id);
+    }
+
+    const after = await getReport6();
+
+    expect(Number(after[10][HEALTH_AUTHORITIES[0].name])).toBe(1);
```
sunghwan-park-fw, line 91: How about using a loop?

apps/api/test/06-report-6.e2e-spec.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  it('Add new job/status to applicant with job/status in another HA to Provincial Health Services', async () => {
+    const before = await getReport6();
+
+    await addJob(applicantId, HEALTH_AUTHORITIES[4].id, 'TestTwoStatusSameApplicant');
+    await addMilestone(applicantId, undefined);
+
+    const after = await getReport6();
+
+    expect(Number(before[10][PHSA])).toBe(0);
```
sunghwan-park-fw, line 105: this test doesn't do anything to make the initial value 0. How about testing just the increase considering that it could fail if we add more milestones to the seed data?

apps/api/test/06-report-6.e2e-spec.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    return body;
+  };
+
+  // add a milestone
+  const addMilestone = async (id: string, status_id?: string, start_date?: string) => {
+    const addStatusUrl = `/ien/${id}/status`;
+    const milestone = getMilestone({
+      status: status_id || RECRUITMENT_STAGE_STATUSES[1].id,
+      job_id: jobTempId || 'WillHaveJobId',
```
sunghwan-park-fw, line 70: I wonder if the milestone could be added even if job_id is set to 'WillHaveJobId'. If not, I'd like to `expect` job_id to be defined.

apps/api/test/report-util.ts
```diff
@@ -67,3 +83,61 @@ export const getEducation = (options?: EducationOptions): NursingEducationDTO =>
     num_years: faker.helpers.arrayElement([1, 2, 3, 4]),
   };
 };
+
+export const HEALTH_AUTHORITIES = [
```
sunghwan-park-fw, line 87: We can get this data by using `Repository` or referencing `Service`

apps/api/test/06-report-6.e2e-spec.ts
coryatwater-FW: Just figured 4 or so would be good instead of running the test for all 9 or 10 HA's

apps/api/test/06-report-6.e2e-spec.ts
coryatwater-FW: hmm don't remember why I put that there, will double check

apps/api/test/report-util.ts
coryatwater-FW: ahh forgot about these

sunghwan-park-fw (approved):

apps/api/test/06-report-6.e2e-spec.ts
```diff
@@ -0,0 +1,176 @@
...
+  const getReport6 = async () => {
+    const { body } = await request(app.getHttpServer()).get(URLS.REPORT6);
+    return body;
+  };
+
+  const addApplicant = async (applicant: IENApplicantCreateUpdateDTO) => {
+    const { body } = await request(app.getHttpServer()).post('/ien').send(applicant);
+    return body;
+  };
+
```
dbayly-freshworks, line 68: Can addJob/addMilestone/addApplicant be moved to the util file and used by other tests?

coryatwater-FW: yes, I moved them in my report 7 test branch, but I think a separate PR to replace all the requests to use these fn would be best

PR #467, opened 2023-02-09
IEN-742 - report 8 test

coryatwater-FW:
add report 8 test


apps/api/test/report-util.ts
```diff
@@ -65,14 +65,15 @@ export const COUNTRY_OF_EDUCATIONS = {
 };
 
 export const RECRUITMENT_STAGE_STATUSES = {
```
sunghwan-park-fw, line 67: I removed this object and added a method to query status id in my pr.

apps/api/test/08-report-8.e2e-spec.ts
```diff
@@ -0,0 +1,131 @@
...
+    await hire(applicantId, 'IHA', dayjs().format('YYYY-MM-DD'));
+
+    await addMilestone(applicantId, '', {
+      status: await getStatusId(STATUS.RECEIVED_WORK_PERMIT),
+      start_date: fiscalDate,
+    });
+    const after = await getReport8();
+
+    // same applicant, should count only highest/ most recent work permit milestone
+    expect(Number(after[1]['total'])).toBe(Number(before[1]['total']) - 1);
```
sunghwan-park-fw, line 108: Why does the second HA's total decrease?

coryatwater-FW: the docs say the report is supposed to take the most recent milestone, assuming to prevent duplicates in the data.  Just an example, if the Received Work Permit milestone is added twice, it won't count it as 2.  Not sure how frequent that would happen, or if it will.  Can remove the test if we deem it unnecessary

coryatwater-FW: I'd have to double check, but don't we take the 'latest' milestone in a few other reports as well

sunghwan-park-fw: Let's say that an applicant got a work permit in 2020.1 and reporting period is 2021.1 ~ 2022.1.
Then, the work permit got reissued in 2023.1. If we count only the latest one, then the applicant is considered as not working during the period.

apps/api/test/08-report-8.e2e-spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  const getReport8 = async () => {
+    const { body } = await request(app.getHttpServer()).get(URLS.REPORT8);
+    return body;
+  };
+
+  it('Add status to fiscal year', async () => {
+    const before = await getReport8();
+
+    const applicant = getApplicant();
+    applicant.registration_date = '2022-06-01';
```
dbayly-freshworks, line 51: Will this break in a year? Would it be possible to use dayjs to have this be 6 months in the past or something like that?

coryatwater-FW: It will not.  That's just the applicants registration date, report 8 is looking for the date of `Received Work Permit (Arrival in Canada)`. A few lines down the start_date is '2021-01-01' so moving forward it should never be included in the report

coryatwater-FW: It will not.  That's just the applicants registration date, report 8 is looking for the date of `Received Work Permit (Arrival in Canada)`. A few lines down the start_date is dayjs(`${lastYear}-06-06`).format('YYYY-MM-DD');so moving forward it should never fail

sunghwan-park-fw (approved):

PR #475, opened 2023-02-14
IEN-784 - add flag to sort old and new process candidates

coryatwater-FW:
add flag to differentiate between old and new process applicants


dbayly-freshworks (commented):
Looks pretty good. 
I'd suggest adding a @BeforeIsnsert function to the ienapplicant entity to populate the field whenever a new user is added.

dbayly-freshworks (approved):

PR #476, opened 2023-02-14
IEN-734 Create test report 4

JGH-EY:
Created four test cases for Report 4. It includes hard coded data in a separate file as well.

The tests mainly cover the adding and removing of milestones for applicants. I had to include a delete function for each test so that Report 5 would not fail.


JGH-EY:
I see the issue, also, need to reformat the tests using the functions Cory created for 5.

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -0,0 +1,160 @@
...
+    const statusIds: string[] = [];
+
+    const { body: before } = await request(app.getHttpServer()).get(URLS.REPORT4);
+
+    // What the report is expected to look like. Used for comparison
+    const expectedResult = before.map((item: { applicants: string; status: string }) => {
+      let result = parseInt(item.applicants) + 1;
+      // Accounts for other two BCCNM Licenses
+      if (
+        item.status === 'Granted provisional licensure' ||
```
sunghwan-park-fw, line 65: How about using `STATUS` enum?

apps/api/test/fixture/reports.ts
```diff
@@ -26,3 +28,61 @@ export const addMilestone = {
...
+];
+export const jobInput: IENApplicantJobCreateUpdateDTO = {
+  ha_pcn: '6ad69443-e3a8-3cbc-8cc9-3a05e5b771e4',
+  job_id: '148593',
+  recruiter_name: 'Test 4',
+};
+
+export type ReportTestStatus = LICENSE_STATUS | BCCNM_STATUS;
+
+export enum LICENSE_STATUS {
```
sunghwan-park-fw, line 48: With enum and constants in `milestone-status.ts`, I can't catch why we need this additional enum.

apps/api/test/report-util.ts
```diff
@@ -103,13 +104,13 @@ export const getIndexOfStatus = (arr: unknown[], compareTo: string) => {
   return arr.findIndex((v: any) => v.status === compareTo);
 };
 
-export const getStatusId = async (status: STATUS): Promise<string> => {
+export const getStatusId = async (status: STATUS | ReportTestStatus): Promise<string> => {
```
sunghwan-park-fw, line 107: shouldn't STATUS be a superset of ReportTestStatus?

JGH-EY: I see what you mean here. Resolving this currently.

apps/api/test/fixture/reports.ts
JGH-EY: We don't need it, you are correct. I am now aware of the other enums and will correct this. Thanks

apps/api/test/04-report-4.e2e-spec.ts
JGH-EY: The hardcoded licensure strings are used here because they don't exist on STATUS. For the withdrawn status, you are correct. I will fix that now.

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -0,0 +1,265 @@
...
+      if (stat === STATUS.APPLIED_TO_NNAS) {
+        result += 2;
+      }
+      return {
+        ...item,
+        applicants: result.toString(),
+      };
+    });
+
+    const job1 = await addJob(applicantIds[5], {
```
sunghwan-park-fw, line 236: A job is relevant to the recruitment milestones.

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -0,0 +1,265 @@
...
+    // Add applicants and give milestones/status
+    for (const status of LIC_REG_STAGE) {
+      // Job id only needed for WITHDREW_FROM_PROGRAM
+      let jobId = '';
+      const applicant = await addApplicant(getApplicant());
+      applicantIds.push(applicant.id);
+
+      if (status === STATUS.WITHDREW_FROM_PROGRAM) {
+        const { id } = await addJob(applicant.id, {
+          ha_pcn: haPcnArray[0],
```
sunghwan-park-fw, line 90: You can refer to the report 6 test using `report-request-util.getHAs`. Also, I wonder if a job is required to add the `withdrew from program` milestone.

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -0,0 +1,265 @@
...
+      }
+      const milestone = await addMilestone(applicant.id, jobId, await getStatus(status));
+      statusIds.push(milestone.id);
+    }
+
+    const { body: after } = await request(app.getHttpServer()).get(URLS.REPORT4);
+    expect(after).toStrictEqual(expectedResult);
+
+    // Resetting milestones to original
+    for (let i = 0; i <= LIC_REG_STAGE.length - 1; i++) {
```
sunghwan-park-fw, line 104: This test may not be affected by other tests. If other tests fail because of applicants and milestones added by this test, why don't we fix them instead of cleaning test data?

JGH-EY: Just to clarify, you mean if other tests fail because of my tests, then I should fix the other tests and not clean the test data in my test? I figured that isolating my test would be best so others don't have to worry about fixing my test in the future. Do you think this will cause issues in the future?

JGH-EY: By other tests I mean tests other than report 4 test

apps/api/test/04-report-4.e2e-spec.ts
JGH-EY: > A job is relevant to the recruitment milestones.

Thanks for the clarification. I misunderstood this.

apps/api/test/04-report-4.e2e-spec.ts
sunghwan-park-fw: I meant that we just need to write tests so that they will not be affected by other tests or will not affect others, but not both. In my opinion, the former is harder than the latter.

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+    const { body: before } = await request(app.getHttpServer()).get(URLS.REPORT4);
+    // What the report is expected to look like. Only looks at old process
+    const expectedResult = reportFourExpectedResult(before, false);
+
+    // Add applicants and give milestones/status
+    for (const status of LIC_REG_STAGE) {
+      // Job id only needed for WITHDREW_FROM_PROGRAM
+      const applicant = await addApplicant(getApplicant({ between: ['2022-05-29', '2022-06-19'] }));
+      applicantIdsOldProcess.push(applicant.id);
+      const milestone = await addMilestone(applicant.id, '', await getStatus(status));
```
coryatwater-FW, line 67: doesn't look like this const is being used

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    const licenseAfter = reportFourNumberOfApplicants(after, STATUS.BCCNM_FULL_LICENCE_LPN, false);
+    // When the applicant gets hired they should be subtracted from the report value
+    expect(licenseAfter).toBe((parseInt(licenseBefore!) - 1).toString());
+  });
+
+  it('Applicant with all 4 licenses and check granted licensure', async () => {
+    const { body: before } = await request(app.getHttpServer()).get(URLS.REPORT4);
+
+    for (const stat of BCCNM_LICENSE) {
```
coryatwater-FW, line 165: maybe change `stat` to `status` to stay consistent with your other loops?

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+    const { body: before } = await request(app.getHttpServer()).get(URLS.REPORT4);
+    const nnasBefore = reportFourNumberOfApplicants(before, STATUS.APPLIED_TO_NNAS, false);
+
+    await addMilestone(applicantIdsOldProcess[5], '', await getStatus(STATUS.RECEIVED_NNAS_REPORT));
+    await addMilestone(applicantIdsOldProcess[6], '', await getStatus(STATUS.SUBMITTED_DOCUMENTS));
+
+    const { body: after } = await request(app.getHttpServer()).get(URLS.REPORT4);
+    const nnasAfter = reportFourNumberOfApplicants(after, STATUS.APPLIED_TO_NNAS, false);
+
+    expect(nnasBefore).toStrictEqual((parseInt(nnasAfter!) - 2).toString());
```
coryatwater-FW, line 198: I prefer `Number` rather than parseInt, it's more strict when converting strings to numbers.  Just in the case we get back a weird string

apps/api/test/fixture/reports.ts
```diff
@@ -1,3 +1,4 @@
+import { STATUS } from '@ien/common/src/enum';
 import { randomUUID } from 'crypto';
```
coryatwater-FW, line 2: we like to do project files at bottom and everything else at top for import organizing

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+  it('Applicant with all 4 licenses and check granted licensure', async () => {
+    const { body: before } = await request(app.getHttpServer()).get(URLS.REPORT4);
+
+    for (const stat of BCCNM_LICENSE) {
+      const licenseBefore = reportFourNumberOfApplicants(before, stat, false);
+      const milestone = await addMilestone(applicantIdsOldProcess[2], '', await getStatus(stat));
+      const { body: afterLicense } = await request(app.getHttpServer()).get(URLS.REPORT4);
+      const licenseAfter = reportFourNumberOfApplicants(afterLicense, stat, false);
+      expect(licenseAfter).toBe((parseInt(licenseBefore!) + 1).toString());
+      await deleteApplicantStatus(applicantIdsOldProcess[2], milestone.id);
```
coryatwater-FW, line 171: is this delete needed still?

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -0,0 +1,200 @@
...
+    })
+      .overrideGuard(AuthGuard)
+      .useValue({ canActivate })
+      .compile();
+
+    app = moduleFixture.createNestApplication();
+    await app.init();
+
+    setApp(app);
+    const { id } = await getHAs();
```
sunghwan-park-fw, line 44: I think `getHAs` returns an array of authorities.

apps/api/test/04-report-4.e2e-spec.ts
sunghwan-park-fw: I vote for `+nnasAfter`

apps/api/test/fixture/reports.ts
```diff
@@ -26,3 +27,10 @@ export const addMilestone = {
   notes: 'Test Note 1',
   reason: '',
 };
+
+// Report 4 data
+export interface ReportFourItem {
```
sunghwan-park-fw, line 32: A fixture is supposed to provide test data. We'd better define returning object types for report endpoints.

apps/api/test/report-util.ts
```diff
@@ -137,3 +138,59 @@ export const getHaId = async (ha: keyof typeof Authorities): Promise<string> =>
...
+ * Gives same results as adding all milestones to different applicants
+ * @param body report 4 initial values
+ * @param isNewProcess look for new process if true, old otherwise
+ * @return expected result for adding every licensing milestone
+ */
+export const reportFourExpectedResult = (body: ReportFourItem[], isNewProcess: boolean) => {
+  const process = isNewProcess ? 'newProcessApplicants' : 'oldProcessApplicants';
+
+  return body.map(
+    (item: { status: string; oldProcessApplicants: string; newProcessApplicants: string }) => {
```
sunghwan-park-fw, line 170: do we need this type definition?

apps/api/test/04-report-4.e2e-spec.ts
coryatwater-FW: or that, I think we use `+` majority of the time anyways, so best to stick with what we're already using (need to update a few reports, been using Number :P)

apps/api/test/fixture/reports.ts
JGH-EY: Yeah, I wasn't sure where to put that. Just so I understand, are you suggesting I make a new file for report types?

sunghwan-park-fw: You can find an example in `api/src/report/types`.

apps/api/test/04-report-4.e2e-spec.ts
sunghwan-park-fw: What I meant is how object deconstruction is possible for an array. We'd better define the return type of `getHAs` function.

sunghwan-park-fw (approved):

PR #477, opened 2023-02-16
IEN-802 refactor report 9 and 10 using intermediate view tables

sunghwan-park-fw:
- create view tables to calculate durations


apps/api/src/migration/1676405526677-CreateReportView.ts
```diff
@@ -0,0 +1,276 @@
...
+      CREATE OR REPLACE VIEW hired_withdrawn_applicant_milestone AS (
+        SELECT
+          hwa.*,
+          (SELECT min(start_date) FROM public.ien_applicant_status_audit asa WHERE asa.applicant_id = hwa.id) as milestone_start_date,
+          (SELECT min(start_date) FROM public.ien_applicant_status_audit asa
+            WHERE asa.applicant_id = hwa.id AND asa.status_id IN (
+              SELECT ias.id FROM ien_applicant_status ias WHERE ias.status IN (
+                'Applied to NNAS',
+                'Submitted Documents (NNAS Application in Review)',
+                'Received NNAS Report'
```
coryatwater-FW, line 34: would we have to re-write this migration again if they were to change the wording in some of the statuses in ATS?

sunghwan-park-fw: I think so. The latest `typeorm` supports `ViewEntity` in which we can define a view query. Unfortunately, it didn't work properly for the version we are using. I'll try it again.

dbayly-freshworks: Would it be possible to use the id of the Milestone instead of the name? That would allow us to change the name of the milestone in future without modifying the reports.

apps/api/src/migration/1676405526677-CreateReportView.ts
```diff
@@ -0,0 +1,278 @@
...
+                )
+              )
+          ) as competition_outcome,
+          (SELECT min(start_date) FROM public.ien_applicant_status_audit asa
+            WHERE asa.applicant_id = hwa.id AND
+              asa.status_id IN (
+                SELECT ias.id FROM ien_applicant_status ias WHERE ias.category = 'BC PNP Process'
+              )
+          ) as immigration,
+          (SELECT min(start_date) FROM public.ien_applicant_status_audit asa
```
dbayly-freshworks, line 139: Nit - This is pretty hard to read. Would it be possible to add a new line between each select statement?

apps/api/src/migration/1676405526677-CreateReportView.ts
```diff
@@ -0,0 +1,278 @@
...
+            WHERE asa.applicant_id = hwa.id AND
+              asa.status_id = (SELECT ias.id FROM ien_applicant_status ias WHERE ias.status = 'Submitted application to BC PNP')   
+          ) as submitted_bc_pnp_application,
+          (SELECT min(start_date) FROM public.ien_applicant_status_audit asa
+            WHERE asa.applicant_id = hwa.id AND
+              asa.status_id = (SELECT ias.id FROM ien_applicant_status ias WHERE ias.status = 'Received Confirmation of Nomination')   
+          ) as received_confirmation_of_nomination,
+          (SELECT min(start_date) FROM public.ien_applicant_status_audit asa
+            WHERE asa.applicant_id = hwa.id AND
```
dbayly-freshworks, line 156: Lots of repetition here. Would it be possible to create a list constant with status and the variable name as an object, and then loop through it? 
e.g const status_list = [ {status : 'Sent First steps document to candidate', name:'sent_first_steps_document'}]

apps/api/src/migration/1676405526677-CreateReportView.ts
sunghwan-park-fw: @dbayly-freshworks 
If we use ID instead of status, it could only work for just one database. I think we need to add a new field like 'abbreviation' or 'code' to differentiate it from status as 'display text'. I am not sure if we should do it in this ticket.

sunghwan-park-fw: I got ViewEntity working. I'll remove the migration.

dbayly-freshworks: The Ids for each are set both in the migration, and by the ATS sync. I think we have have some level of confidence that the ids will be the same across environments. I have a JSON file that has all the ids if thats useful.

sunghwan-park-fw: ViewEntity doesn't generate a temporary table with `expression` cause we don't sync schema. so, we need the migration to create view tables. I defined a constant map of status and its id and made the long query more readable.

apps/api/src/report/types/duration-entry.ts
```diff
@@ -0,0 +1,8 @@
+export interface DurationEntry {
+  stage?: string;
+  milestone?: string;
+  field: string;
+  Mean?: number;
+  Mode?: number;
+  Median?: number;
```
coryatwater-FW, line 7: how come these are capitalized?

apps/api/src/report/report.service.ts
```diff
@@ -276,22 +287,60 @@ export class ReportService {
...
-      `getAverageTimeOfMilestones: query completed a total of ${data.length} record returns`,
-    );
-    return data;
+    const licRecDurations = await getRepository(MilestoneDurationEntity).find({
+      where: [{ hired_date: LessThanOrEqual(to) }, { withdraw_date: LessThanOrEqual(to) }],
+    });
+
+    // the end date of immigration milestones should be limited by 'to' date instead of hired date
+    // because immigration stage comes after recruitment
+    const immigrationDurations = await getManager().query(`
```
coryatwater-FW, line 296: more so just brainstorming, but this query is duplicated in the `CREATE VIEW` as part of the `milestone_duration` too, with just the `to` date replacing the `CURRENT_DATE`.  Wonder if we can reduce that?  Just leaving this here  while I continue thinking

apps/api/src/report/constants/milestone-id.ts
```diff
@@ -0,0 +1,58 @@
+import { STATUS } from '@ien/common';
+
+export const STATUS_ID: Record<STATUS, string> = {
+  [STATUS.APPLIED_TO_BCCNM]: '8a9b0d13-f5d7-4be3-8d38-11e5459f9e9a',
+  [STATUS.APPLIED_TO_NNAS]: 'e68f902c-8440-4a9c-a05f-2765301de800',
```
coryatwater-FW, line 5: can we use getStatusMap() instead of this?  It returns the statuses in the same format `'Candidate Withdrew from Competition': '3fd4f2b0-5151-d7c8-6bbc-3a0601b5e1b0'`

sunghwan-park-fw: Good catch! I feel trapped in a labyrinth.

apps/api/src/report/report.service.ts
sunghwan-park-fw: That's what I wanted but I couldn't find a way.

apps/api/src/report/types/duration-entry.ts
sunghwan-park-fw: I don't remember exactly. My guess is it's to avoid selectively capitalizing field names when generating spreadsheets.

apps/api/src/report/entity/hired-applicant-milestone.entity.ts
```diff
@@ -0,0 +1,389 @@
...
+        (
+          SELECT
+            min(asa.start_date) AS min
+          FROM
+            ien_applicant_status_audit asa
+          WHERE
+            asa.applicant_id = hwa.id
+            AND (asa.status_id = ANY (ARRAY['e68f902c-8440-4a9c-a05f-2765301de800'::uuid,
+            '897156c7-958c-4ec0-879a-ed4943af7b72'::uuid,
+            '20268c6e-145e-48cd-889a-2346985db957'::uuid]))
```
coryatwater-FW, line 33: do these all need to be hardcoded?

apps/api/src/report/entity/milestone-duration.entity.ts
```diff
@@ -0,0 +1,259 @@
+import { ViewColumn, ViewEntity } from 'typeorm';
+import { HiredApplicantMilestoneEntity } from './hired-applicant-milestone.entity';
+
+// the expression is given by reference
+// because we don't sync schema and the view table should have been created by a migration
```
coryatwater-FW, line 5: could you expand on this? Is this why we need the full query in both `expression` and the `migration`?

apps/api/test/10-report-10.e2e-spec.ts
```diff
@@ -112,36 +117,37 @@ describe('Report 10 - Average Amount of Time with Each Milestone in Stakeholder
...
-      await addMilestone(id, '', await getStatus(STATUS.SUBMITTED_DOCUMENTS, submittedDoc));
+      await addMilestone(id, '', await getStatus(STATUS.SUBMITTED_DOCUMENTS, submittedDocDate));
+      await addMilestone(id, '', await getStatus(STATUS.RECEIVED_WORK_PERMIT, immigratedDate));
     }
 
     report = await getReport(to);
 
     // check APPLIED_TO_NNAS stats
     let { Mean, Mode, Median } = report.find(s => s.milestone === STATUS.APPLIED_TO_NNAS) || {};
     expect(Mean).toBe(round(mean(appliedToNNAS), 2));
```
coryatwater-FW, line 136: you use `round` here but `floor` in `getDurationStats` fn

apps/api/src/report/entity/hired-applicant-milestone.entity.ts
sunghwan-park-fw: It's generated from the view table and I added it as a comment for reference.
To map IDs dynamically, we need to use a function with a param of `datasource` and write typeorm query. I just didn't want to bother the codes that won't be run.

apps/api/src/report/report.service.ts
```diff
@@ -269,46 +272,120 @@ export class ReportService {
...
   async getAverageTimeWithEachStakeholderGroup(statuses: Record<string, string>, t: string) {
     const { to } = this.captureFromTo('', t);
     this.logger.log(`getAverageTimeWithEachStakeholderGroup: apply filter till ${to} date)`);
 
-    const entityManager = getManager();
-    const data = await entityManager.query(
-      this.reportUtilService.averageTimeWithEachStakeholderGroupQuery(statuses, to),
-    );
-    this.logger.log(
-      `getAverageTimeWithEachStakeholderGroup: query completed a total of ${data.length} record returns`,
```
coryatwater-FW, line 283: any chance of keeping these logs in report 9 and 10?

sunghwan-park-fw: Because we can view the list of applicants with relevant milestones by querying on the view, I don't think logging of intermediate queries is required.

coryatwater-FW (approved):
LGTM

PR #478, opened 2023-02-16
IEN-798 - migration to remove old unused tables

coryatwater-FW:
removes unused tables created during initial project setup

Note: only `form` and `submission` were found in dev, test and prod.  The other applicant tables seem to just be outliers in some local environments

### DIFF
--- a/apps/api/src/migration/1676577091787-RemoveOldUnusedTables.ts
+++ b/apps/api/src/migration/1676577091787-RemoveOldUnusedTables.ts
@@ -5,10 +5,10 @@ export class RemoveOldUnusedTables1676577091787 implements MigrationInterface {
     await queryRunner.query(`
       DROP TABLE IF EXISTS form;
       DROP TABLE IF EXISTS submission;
-      DROP TABLE IF EXISTS applicant_status;
-      DROP TABLE IF EXISTS applicants;
-      DROP TABLE IF EXISTS applicant_status_audit;
-      DROP TABLE IF EXISTS applicant_audit;
+      DROP TABLE IF EXISTS applicant_status CASCADE;
+      DROP TABLE IF EXISTS applicants CASCADE;
+      DROP TABLE IF EXISTS applicant_status_audit CASCADE;
+      DROP TABLE IF EXISTS applicant_audit CASCADE;
     `);
   }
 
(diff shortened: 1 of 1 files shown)


sunghwan-park-fw (approved):

apps/api/src/migration/1676577091787-RemoveOldUnusedTables.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class RemoveOldUnusedTables1676577091787 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      DROP TABLE IF EXISTS form;
+      DROP TABLE IF EXISTS submission;
+      DROP TABLE IF EXISTS applicant_status;
+      DROP TABLE IF EXISTS applicants;
+      DROP TABLE IF EXISTS applicant_status_audit;
+      DROP TABLE IF EXISTS applicant_audit;
```
dbayly-freshworks, line 11: I am a bit concerned about the applicant/applicant_status/applicant_audit tables having relationships between them, preventing a delete. Would you be able to add "CASCADE" here?

apps/api/src/migration/1676577091787-RemoveOldUnusedTables.ts
```diff
@@ -0,0 +1,18 @@
...
+    `);
+  }
+
+  public async down(): Promise<void> {
+    // no rollback with this migration
```
dbayly-freshworks, line 16: Do we have a migration with these initial tables? If so, it would be a nice to have to be able to add them back in. 
Totally a nitpick though.

apps/api/src/migration/1676577091787-RemoveOldUnusedTables.ts
coryatwater-FW: hmm, wonder if it's better I just take these 4 out?  I don't know about these tables relations, do you recall if they were referencing any existing/ in use tables?

apps/api/src/migration/1676577091787-RemoveOldUnusedTables.ts
coryatwater-FW: I can't find any reference to these tables in the project

apps/api/src/migration/1676577091787-RemoveOldUnusedTables.ts
coryatwater-FW: nevermind, found the commit the migrations were removed, ref [here](https://github.com/bcgov/internationally-educated-nurses/pull/33/files#:~:text=apps/api/src/migration/1645638796025%2DApplicantAndStatus.ts), looks like they only reference each other

PR #480, opened 2023-02-17
IEN-800 - fix duplicate milestone in test env

coryatwater-FW:
The test environment did not have the unique index `unique_applicant_status_date_job`, can only assume there were already duplicates when it tried creating it?.  Local, dev and prod all have it and as far as I can tell no migration removing it.  Wrote a small migration just to prevent any further issues with it not being created

**Already removed the duplicate entries in Test


coryatwater-FW:
No longer eneded, will do it manually on test DB

PR #481, opened 2023-02-21
IEN-785 - add new column for new process applicants

coryatwater-FW:
added new column for new process applicants


dbayly-freshworks (approved):
Minus cypress issues this looks good to me.

PR #485, opened 2023-02-27
IEN-806 - create s3 bucket to deploy api lambda

coryatwater-FW:


terraform/api.tf
```diff
@@ -1,14 +1,55 @@
...
+        }
+      }
+    ]
+  })
+}
+
+resource "aws_s3_bucket_object" "api_lambda" {
+  bucket = aws_s3_bucket.api.bucket
+  key    = "api-lambda-s3"
+  source = "./build/empty_lambda.zip"
```
sunghwan-park-fw, line 37: How about specifying api.zip, supposedly built just before running terraform?

Makefile
```diff
@@ -297,6 +299,7 @@ runs:
 
 sync-app:
 	aws s3 sync ./terraform/build/app s3://$(APP_SRC_BUCKET) --delete
+	aws s3 sync ./terraform/build/api s3://$(API_SRC_BUCKET) --delete
```
sunghwan-park-fw, line 302: The task is `sync-app` and it shouldn't delete API.
Unlike app, api.zip is just one file so I doubt we need to delete it.
Anyway, we need to define how to upload it. We've done it by running `aws lambda update-function-code`. Now we might need two steps, first upload to s3 and update the function from s3.

terraform/api.tf
coryatwater-FW: ah I tried that, says error file not found, troubleshooting that now

terraform/iam.tf
```diff
@@ -17,6 +17,31 @@ resource "aws_iam_role" "lambda" {
   })
 }
 
+resource "aws_s3_bucket_policy" "api_lambda" {
```
sunghwan-park-fw, line 20: With aws_iam_role_policy_attachment.lambda_s3, do we still need this?

Makefile
```diff
@@ -298,16 +300,20 @@ runs:
 sync-app:
 	aws s3 sync ./terraform/build/app s3://$(APP_SRC_BUCKET) --delete
 
+upload-sync-api:
+	aws s3 sync ./terraform/build/api s3://$(API_SRC_BUCKET)
```
sunghwan-park-fw, line 304: Do we need this?

dbayly-freshworks (approved):

PR #490, opened 2023-03-01
Generate milestones

sunghwan-park-fw:
- Skip applicants with a job
- Skip applicants registered before 2022-06-01
- Do not create recruitment outcome milestones except job offer accepted
- Each milestone's duration is randomly selected between 1 and 10


packages/scripts/src/generate-milestones.ts
```diff
@@ -0,0 +1,175 @@
...
+        Accept: 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded',
+      },
+    },
+  );
+  return resp.data.access_token;
+};
+
+const getApplicants = async () => {
+  const resp = await axios.get('/ien?from=2022-01-01');
```
coryatwater-FW, line 47: does this endpoint support `from` param ?

packages/scripts/src/generate-milestones.ts
```diff
@@ -0,0 +1,175 @@
...
+
+setToken()
+  .then(async () => {
+    const [applicants, count] = await getApplicants();
+    const statuses = await getMilestones();
+    for (let i = 0; i < count; i++) {
+      await checkToken(start);
+
+      const applicant = applicants[i];
+      if (dayjs(applicant.registration_date).isAfter('2022-06-01')) {
```
coryatwater-FW, line 161: comments say `Skip applicants registered before 2022-06-01` should this be isBefore?

sunghwan-park-fw: It generates milestones with random duration, and if the applicant has been registered recently, a milestone date could be a future date.

packages/scripts/src/generate-milestones.ts
```diff
@@ -0,0 +1,175 @@
...
+
+      await generateMilestones(applicant.id, statuses);
+      console.log(`${i}/${count} applicant processed <- ${applicant.id}`);
+    }
+  })
+  .catch(e => {
+    console.log(e);
+  })
+  .finally(() => {
+    console.log('generated milestones');
```
coryatwater-FW, line 174: I assume console logs are fine in the script folder? no need for `Logger`?

sunghwan-park-fw: Yes. I added a comment disabling eslint no-console rule at the top.

PR #498, opened 2023-03-06
IEN-816 call each report endpoint one by one

sunghwan-park-fw:


apps/api/src/report/report.service.ts
```diff
@@ -439,7 +447,14 @@ export class ReportService {
       this.getAverageTimeWithEachStakeholderGroup(statuses, to).then(report9 => ({ report9 })),
       this.getAverageTimeOfMilestones(to).then(report10 => ({ report10 })),
     ];
-    const report = await Promise.all(promises);
+    const report = await Promise.all(
```
coryatwater-FW, line 450: is this function used anymore?

sunghwan-park-fw: `/reports/applicant` endpoint uses this function and we can use the endpoint to get all reports at once by postman.

coryatwater-FW (approved):

PR #501, opened 2023-03-07
IEN-819 insert milestones of new applicants

sunghwan-park-fw:
- insert milestones of new applicants
- do not insert the same milestone without start_date for an applicant

### DIFF
--- a/apps/api/src/applicant/external-api.service.ts
+++ b/apps/api/src/applicant/external-api.service.ts
@@ -533,12 +533,11 @@ export class ExternalAPIService {
    * @returns object that use in upsert milestone/status
    */
   async mapMilestones(data: any, mappedApplicantList: string[]) {
-    const { users } = await this.getApplicantMasterData();
-    const savedApplicants = await this.ienapplicantRepository.findByIds(mappedApplicantList);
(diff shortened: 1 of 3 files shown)


apps/api/src/applicant/external-api.service.ts
```diff
@@ -533,12 +533,11 @@ export class ExternalAPIService {
    * @returns object that use in upsert milestone/status
    */
   async mapMilestones(data: any, mappedApplicantList: string[]) {
-    const { users } = await this.getApplicantMasterData();
-    const savedApplicants = await this.ienapplicantRepository.findByIds(mappedApplicantList);
```
coryatwater-FW, line 537: we don't need this anymore?  If not, cant we check `if (!mappedApplicantList.length)` prior to calling this function?

dbayly-freshworks (approved):

apps/api/src/applicant/external-api.service.ts
sunghwan-park-fw: mappedApplicantList already contains the upserted applicant IDs and `findByIds` returns nothing. I guess it's because of `transaction`.

PR #503, opened 2023-03-08
IEN -818 - add cron job, add report_cache migration, modify report 4 endpoint

coryatwater-FW:
add cron job, add report_cache migration, change report 4 to select from cached table depending on period

### DIFF
--- a/apps/api/src/cachereportfour.ts
+++ b/apps/api/src/cachereportfour.ts
@@ -1,0 +1,31 @@
+import { NestFactory } from '@nestjs/core';
+import { Context, Handler } from 'aws-lambda';
+import { AppModule } from './app.module';
+import { AppLogger } from './common/logger.service';
+import { ReportService } from './report/report.service';
+
+/**
+ * Design this function to trigger existing NestJs appliation services without Api-Getway
+ * All the schedule and backgroud job trigger will be added here.
+ * This handler will cache the report 4 data for every existing period
+ */
+export const handler: Handler = async (event, context: Context) => {
+  const app = await NestFactory.createApplicationContext(AppModule);
+  const reportService = app.get(ReportService);
+
+  const appLogger = app.get(AppLogger);
+  appLogger.log({ event });
+  appLogger.log({ context });
+  try {
+    if (event.hasOwnProperty('path')) {
+      if (event.path === 'cache-report-four') {
+        appLogger.log('Start caching report 4...');
+
+        await reportService.saveReportFourCache();
+      }
+    }
+  } catch (e) {
+    appLogger.error(e);
+  }
+  appLogger.log('...end caching report 4');
+};
--- a/apps/api/src/cachereportfour.ts
+++ b/apps/api/src/cachereportfour.ts
@@ -17,12 +17,10 @@ export const handler: Handler = async (event, context: Context) => {
   appLogger.log({ event });
   appLogger.log({ context });
   try {
-    if (event.hasOwnProperty('path')) {
-      if (event.path === 'cache-report-four') {
-        appLogger.log('Start caching report 4...');
+    if (event.path === 'cache-report-four') {
+      appLogger.log('Start caching report 4...');
 
-        await reportService.saveReportFourCache();
-      }
+      await reportService.saveReportFourCache();
     }
   } catch (e) {
     appLogger.error(e);
--- a/apps/api/src/ormconfig.ts
+++ b/apps/api/src/ormconfig.ts
@@ -9,7 +9,7 @@ const config: PostgresConnectionOptions = {
   host: process.env.POSTGRES_HOST,
   type: 'postgres',
   port: +(process.env.PORTGRES_PORT || 5432),
-  connectTimeoutMS: process.env.NODE_ENV === 'test' ? 30000 : 5000,
+  connectTimeoutMS: process.env.NODE_ENV === 'test' ? 30000 : 40000,
   username: process.env.POSTGRES_USERNAME || 'freshworks',
   password: process.env.POSTGRES_PASSWORD,
   database: process.env.POSTGRES_DATABASE || 'ien',
--- a/apps/api/src/ormconfig.ts
+++ b/apps/api/src/ormconfig.ts
@@ -9,7 +9,7 @@ const config: PostgresConnectionOptions = {
   host: process.env.POSTGRES_HOST,
   type: 'postgres',
   port: +(process.env.PORTGRES_PORT || 5432),
-  connectTimeoutMS: process.env.NODE_ENV === 'test' ? 30000 : 40000,
+  connectTimeoutMS: +(process.env.POSTGRES_TIMEOUT || 30000),
   username: process.env.POSTGRES_USERNAME || 'freshworks',
   password: process.env.POSTGRES_PASSWORD,
   database: process.env.POSTGRES_DATABASE || 'ien',
--- a/apps/api/src/report/report.controller.ts
+++ b/apps/api/src/report/report.controller.ts
@@ -59,8 +59,18 @@ export class ReportController {
   @ApiOperation({ summary: 'Report 4: licensing stage applicants' })
   @Get('/applicant/licensing-stage')
   @AllowAccess(Access.REPORTING)
-  async getLicensingStageApplicants(@Query('period') period: number): Promise<object[]> {
-    return this.reportService.getLicensingStageApplicants(period);
+  async getLicensingStageApplicants(
+    @Query('period') period: number,
+    @Query('from') from: string,
+    @Query('to') to: string,
+    @Query('testReport') testReport = false,
+  ): Promise<object[]> {
+    // cached data relies on cron job set to run at 12AM PST
+    // testReport param is used to invalidate the cache and use current data for testing
+    if (testReport || !period) {
+      return this.reportService.splitReportFourNewOldProcess(from, to);
+    }
+    return this.reportService.getLicensingStageApplicants(period, from, to);
   }
 
   @ApiOperation({ summary: 'Report 5: Applicants eligible for job search' })
--- a/apps/api/src/report/report.controller.ts
+++ b/apps/api/src/report/report.controller.ts
@@ -70,7 +70,7 @@ export class ReportController {
     if (noCache || !period) {
       return this.reportService.splitReportFourNewOldProcess(from, to);
     }
-    return this.reportService.getLicensingStageApplicants(period, from, to);
+    return this.reportService.getLicensingStageApplicants(period);
   }
 
   @ApiOperation({ summary: 'Report 5: Applicants eligible for job search' })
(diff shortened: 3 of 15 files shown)


apps/api/src/cachereportfour.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const handler: Handler = async (event, context: Context) => {
+  const app = await NestFactory.createApplicationContext(AppModule);
+  const reportService = app.get(ReportService);
+
+  const appLogger = app.get(AppLogger);
+  appLogger.log({ event });
+  appLogger.log({ context });
+  try {
+    if (event.hasOwnProperty('path')) {
```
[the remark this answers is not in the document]
coryatwater-FW: I'm honestly not even sure what this is, that's why it's still in draft haha

apps/api/src/cachereportfour.ts
```diff
@@ -0,0 +1,31 @@
...
+import { AppModule } from './app.module';
+import { AppLogger } from './common/logger.service';
+import { ReportService } from './report/report.service';
+
+/**
+ * Design this function to trigger existing NestJs appliation services without Api-Getway
+ * All the schedule and backgroud job trigger will be added here.
+ * This handler will cache the report 4 data for every existing period
+ */
+export const handler: Handler = async (event, context: Context) => {
```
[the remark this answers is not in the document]
coryatwater-FW: hmm, I get that point, I split it up to keep it a bit more modular as I felt reports could become quite heavy and it'd be better to keep it separate.  thoughts @dbayly-freshworks @JGH-EY

apps/api/src/report/report.controller.ts
```diff
@@ -55,12 +59,8 @@ export class ReportController {
   @ApiOperation({ summary: 'Report 4: licensing stage applicants' })
   @Get('/applicant/licensing-stage')
   @AllowAccess(Access.REPORTING)
-  async getLicensingStageApplicants(
-    @Query('from') from: string,
-    @Query('to') to: string,
-  ): Promise<object[]> {
-    const statuses = await this.reportService.getStatusMap();
-    return this.reportService.getLicensingStageApplicants(statuses, from, to);
+  async getLicensingStageApplicants(@Query('period') period: number): Promise<object[]> {
```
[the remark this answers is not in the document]
coryatwater-FW: what do you mean?  I figured the data would all be outdated regardless, since the sync only runs at certain times, so if the caching happens shortly after the sync, the data should be up to date, no?  that's my theory at least

apps/api/src/cachereportfour.ts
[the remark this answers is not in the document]
sunghwan-park-fw: I think we don't need this, at least in our case.

apps/api/src/report/report.controller.ts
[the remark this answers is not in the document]
sunghwan-park-fw: What if we want to see the up-to-date report instead of the cached one just after updating new data?
I am sure Akhil wants to do that.

apps/api/src/report/report.service.ts
```diff
@@ -148,20 +150,86 @@ export class ReportService {
...
+    const report_number = 4;
+    const report = await getRepository(ReportCacheEntity)
+      .createQueryBuilder('rce')
+      .where('rce.report_period = :period', { period })
+      .andWhere('rce.report_number = :report_number', { report_number })
+      .orderBy('rce.created_date', 'DESC')
+      .getOne();
+
+    if (!report) {
+      return [];
```
sunghwan-park-fw, line 168: In the normal caching process, shouldn't we try to fetch data if the cache is not found?

apps/api/src/report/report.controller.ts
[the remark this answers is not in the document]
coryatwater-FW: ahhh, right, forgot about postman testing!

apps/api/src/ormconfig.ts
```diff
@@ -9,7 +9,7 @@ const config: PostgresConnectionOptions = {
   host: process.env.POSTGRES_HOST,
   type: 'postgres',
   port: +(process.env.PORTGRES_PORT || 5432),
-  connectTimeoutMS: process.env.NODE_ENV === 'test' ? 30000 : 5000,
+  connectTimeoutMS: process.env.NODE_ENV === 'test' ? 30000 : 40000,
```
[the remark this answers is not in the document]
coryatwater-FW: yeah, I'll change it to just 30000 and remove the ternary, was just a random test I was doing

apps/api/src/report/report.service.ts
coryatwater-FW: yep, I will fix this when I add in the option to invalidate the cache

apps/api/src/cachereportfour.ts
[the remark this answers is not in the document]
JGH-EY: Not sure what its really for as well. Its not throwing an error or anything.

apps/api/src/cachereportfour.ts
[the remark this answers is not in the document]
JGH-EY: I can see both your points. If we are expecting a bunch of different paths to check, I think Cory's idea would be more readable and easier to maintain. In my opinion, really depends on the number of variations of event.path to compare.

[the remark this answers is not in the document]
sunghwan-park-fw: One path might be enough to cache all reports we need.

apps/api/src/cachereportfour.ts
```diff
@@ -0,0 +1,29 @@
...
+  const reportService = app.get(ReportService);
+
+  const appLogger = app.get(AppLogger);
+  appLogger.log({ event });
+  appLogger.log({ context });
+  try {
+    if (event.path === 'cache-report-four') {
+      appLogger.log('Start caching report 4...');
+
+      await reportService.saveReportFourCache();
```
sunghwan-park-fw, line 23: How about changing `saveReportFourCache` to `updateReportCache`? We could add reports to that function later if we need to cache more reports without modifying the lambda.

terraform/report_lambda_function.tf
```diff
@@ -0,0 +1,64 @@
+resource "aws_lambda_function" "CacheReportFour" {
+  description      = "Trigger Caching of Report 4"
+  function_name    = local.cache_report_four_lambda_name
+  role             = aws_iam_role.lambda.arn
+  runtime          = "nodejs14.x"
+  filename         = "./build/empty_lambda.zip"
+  source_code_hash = filebase64sha256("./build/empty_lambda.zip")
+  handler          = "api/cachereportfour.handler"
+  memory_size      = var.function_memory_mb
+  timeout          = 300
```
sunghwan-park-fw, line 10: Considering why we need to cache, don't we need to increase the timeout?

terraform/report_lambda_function.tf
```diff
@@ -0,0 +1,64 @@
...
+      POSTGRES_HOST     = aws_rds_cluster.pgsql.endpoint
+      POSTGRES_DATABASE = aws_rds_cluster.pgsql.database_name
+    }
+  }
+}
+
+resource "aws_cloudwatch_event_rule" "cache_report_four" {
+  name                = local.cache_report_four_lambda_name
+  description         = "7:00AM UTC - 12:00AM PST Every day"
+  schedule_expression = "cron(0 7 * * ? *)"
```
sunghwan-park-fw, line 49: Applicants are synced at 8. If we update the report cache before it, the report will be behind a day.

terraform/report_lambda_function.tf
sunghwan-park-fw: Oh, it's 300 seconds.

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -36,19 +57,43 @@ describe('Report 4 - Number of IEN registrants in the licensing stage', () => {
     setApp(app);
     let temp = await getHAs();
     haPcn = temp[0].id;
+
+    reportService = moduleFixture.get<ReportService>(ReportService);
+
+    const { body: report } = await request(app.getHttpServer()).get(URLS.REPORT1);
+
+    // hard set period ot be current period moving forward
```
dbayly-freshworks, line 65: Typo

dbayly-freshworks (approved):
One nit, otherwise, I think this looks ok. I don't see any red flags. 

Before merging this in, I would recommend that we pull down a report on dev. 
If the report 4 results match pre/post push, I think we should be good to go.

apps/api/src/report/report.service.ts
```diff
@@ -148,14 +150,78 @@ export class ReportService {
...
+  async getLicensingStageApplicants(period: number, from: string, to: string) {
+    const report_number = 4;
+    const report = await getRepository(ReportCacheEntity)
+      .createQueryBuilder('rce')
+      .where('rce.report_period = :period', { period })
+      .andWhere('rce.report_number = :report_number', { report_number })
+      .orderBy('rce.created_date', 'DESC')
+      .getOne();
+
+    if (!report) {
```
sunghwan-park-fw, line 167: We could save the cache before returning.

apps/api/src/report/report.controller.ts
```diff
@@ -56,11 +60,17 @@ export class ReportController {
   @Get('/applicant/licensing-stage')
   @AllowAccess(Access.REPORTING)
   async getLicensingStageApplicants(
+    @Query('period') period: number,
     @Query('from') from: string,
     @Query('to') to: string,
+    @Query('testReport') testReport = false,
```
sunghwan-park-fw, line 66: This name is a little vague. It could be more specific like 'no-cache'

terraform/report_lambda_function.tf
```diff
@@ -0,0 +1,64 @@
...
+      POSTGRES_PASSWORD = data.aws_ssm_parameter.postgres_password.value
+      POSTGRES_HOST     = aws_rds_cluster.pgsql.endpoint
+      POSTGRES_DATABASE = aws_rds_cluster.pgsql.database_name
+    }
+  }
+}
+
+resource "aws_cloudwatch_event_rule" "cache_reports" {
+  name                = local.cache_reports_lambda_name
+  description         = "9:00AM UTC - 2:00AM PST Every day"
```
JGH-EY, line 48: I think this is supposed to be 10 AM UTC for it to be 2:00AM PST

apps/api/src/report/report.controller.ts
sunghwan-park-fw: Period and from/to seem conflicting with each other. What if `from` and `to` are not equal to the period's start and end?

coryatwater-FW: it's mainly for the no-cache call, I need the from and to for it, any ideas?

coryatwater-FW: AH, after having slept on it, I get what you mean... will fix

PR #504, opened 2023-03-09
Ien 817 optimize report four

dbayly-freshworks:


apps/api/src/report/constants/report-four-format.ts
```diff
@@ -0,0 +1,28 @@
+export const REPORT_FOUR_STEPS = [
```
JGH-EY, line 1: More of a clarification comment, are these in order of how the report is collecting the data for the report?

JGH-EY: Also, not 100% sure on this. Should this be in camel case?

JGH-EY: It looks great so far, I really like your idea with the order here.

apps/api/src/report/constants/report-four-format.ts
```diff
@@ -0,0 +1,28 @@
...
+  'Applied to NNAS',
+  'Applied to BCCNM',
+  'Completed English Language Requirement',
+  'Referred to NCAS',
+  'Completed NCAS',
+  'Completed Additional Education',
+  'NCLEX - Written',
+  'NCLEX - Passed',
+  'BCCNM Full Licence LPN',
```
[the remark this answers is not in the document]
sunghwan-park-fw: This order is based on the old process. I guess that the milestones from ATS don't match with the new process yet.

apps/api/src/report/constants/report-four-format.ts
```diff
@@ -0,0 +1,28 @@
+export const REPORT_FOUR_STEPS = [
+  'Applied to NNAS',
```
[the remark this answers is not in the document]
dbayly-freshworks: Yup, I will fix this

apps/api/src/report/constants/report-four-format.ts
```diff
@@ -0,0 +1,31 @@
+import { STATUS } from '@ien/common';
+
+export const REPORT_FOUR_STEPS = [
```
coryatwater-FW, line 3: this is the same as `LIC_REG_STAGE` no? couldn't we just re-order that to match this? doesn't look like it's currently being used for anything that requires a specific order

apps/api/src/report/report.service.ts
```diff
@@ -221,37 +221,149 @@ export class ReportService {
...
-    const entityManager = getManager();
-
     const { from, to } = this.captureFromTo(f, t);
 
     this.logger.log(`getLicensingStageApplicants: Apply date filter from (${from}) and to (${to})`);
+    const connection = getConnection();
+    const mappedStatuses = LIC_REG_STAGE.map(licensingStatus => {
+      return statuses[licensingStatus];
+    });
+    const mappedStatusesString = mappedStatuses.map(status => `'${status}'`).join(',');
```
coryatwater-FW, line 231: could we do this in one line? as in tack on .join to the `mappedStatuses` map and just wrap the return like ``'${statuses[licensingStatus]}'`

apps/api/src/report/report.service.ts
```diff
@@ -221,37 +221,149 @@ export class ReportService {
...
+  }
+  mapReportFourResults(
+    statuses: Record<string, string>,
+    oldProcess: { status_id: string; count: string }[],
+    newProcess: { status_id: string; count: string }[],
+    licenceResults: any,
+    withdrawnOld: string,
+    withdrawnNew: string,
+  ) {
+    let report = REPORT_FOUR_STEPS.map((step: string) => {
```
coryatwater-FW, line 267: `const`?

apps/api/src/report/constants/report-four-format.ts
dbayly-freshworks: That makes sense, I'll give it a shot.

apps/api/src/report/report.service.ts
```diff
@@ -221,37 +221,149 @@ export class ReportService {
...
+      switch (step) {
+        case 'Granted full licensure':
+          return {
+            status: step,
+            oldProcessApplicants: licenceResults.oldFullLicence[0]?.count || '0',
+            newProcessApplicants: licenceResults.newFullLicence[0]?.count || '0',
+          };
```
coryatwater-FW, line 274: wonder if it's worth just creating a function to return this object to reduce the repetitiveness

apps/api/src/report/report.service.ts
```diff
@@ -221,37 +221,149 @@ export class ReportService {
...
-    const oldProcess = await entityManager.query(
-      this.reportUtilService.licensingStageApplicantsQuery(statuses, from, to),
-    );
+        case 'Granted provisional licensure':
+          return {
+            status: step,
+            oldProcessApplicants: licenceResults.oldProvisionalLicence[0]?.count || '0',
+            newProcessApplicants: licenceResults.newProvisionalLicence[0]?.count || '0',
+          };
+        case 'Applied to NNAS':
```
coryatwater-FW, line 282: could probably use `STATUS` for these ones

apps/api/src/report/report.service.ts
```diff
@@ -221,37 +221,149 @@ export class ReportService {
...
+              newProcess.find(value => value.status_id === statuses[step])?.count || '0',
+          };
+      }
+    });
+    return report;
+  }
+  findAndSumStatusCounts(list: { status_id: string; count: string }[], acceptedIds: string[]) {
+    let count = 0;
+    acceptedIds.forEach(id => {
+      count += parseInt(
```
coryatwater-FW, line 345: `+` instead of parseInt? only because we've been using it throughout the project already

apps/api/src/report/report.util.service.ts
```diff
@@ -215,6 +215,120 @@ export class ReportUtilService {
...
+      AND start_date::date <= '${to}' AND 
+      NOT EXISTS (
+        SELECT *
+        FROM "ien_applicant_status_audit" AS T2
+        WHERE 
+          T2.applicant_id = status_audit.applicant_id
+          AND 
+            (
+              T2.start_date > status_audit.start_date 
+              OR T2.status_id IN ('f84a4167-a636-4b21-977c-f11aefc486af', '70b1f5f1-1a0d-ef71-42ea-3a0601b46bc2')
```
coryatwater-FW, line 257: we should replace these with `getStatusMap()`, it returns the status ids

apps/api/src/report/report.util.service.ts
```diff
@@ -215,6 +215,120 @@ export class ReportUtilService {
...
+    return `
+      SELECT 
+        count(distinct status_audit.applicant_id)
+      FROM 
+        "ien_applicant_status_audit" "status_audit" 
+      LEFT JOIN 
+        ien_applicants applicant ON applicant.id = status_audit.applicant_id 
+      WHERE 
+        applicant.new_bccnm_process IS NOT NULL AND applicant.new_bccnm_process = ${
+          BCCNM_NEW_PROCESS ? 'TRUE' : 'FALSE'
```
coryatwater-FW, line 279: same boolean question here

apps/api/src/report/report.service.ts
```diff
@@ -221,37 +221,149 @@ export class ReportService {
...
     const { from, to } = this.captureFromTo(f, t);
 
     this.logger.log(`getLicensingStageApplicants: Apply date filter from (${from}) and to (${to})`);
+    const connection = getConnection();
+    const mappedStatuses = LIC_REG_STAGE.map(licensingStatus => {
+      return statuses[licensingStatus];
+    });
+    const mappedStatusesString = mappedStatuses.map(status => `'${status}'`).join(',');
+    try {
+      const oldProcess = await connection.query(
```
JGH-EY, line 233: I wonder is using an await Promise.all would be useful here? Quite a few await calls in this section.

apps/api/src/report/report.service.ts
JGH-EY: Not sure about this one, but would the type for step here be STATUS?

apps/api/src/report/constants/report-four-format.ts
JGH-EY: I would leave a comment that the order should not be changed on this. Just in case

apps/api/src/report/report.util.service.ts
```diff
@@ -215,6 +215,120 @@ export class ReportUtilService {
...
+    LEFT JOIN 
+      ien_applicants applicant ON applicant.id = status_audit.applicant_id 
+    WHERE 
+      applicant.new_bccnm_process IS NOT NULL AND applicant.new_bccnm_process = ${
+        BCCNM_NEW_PROCESS ? 'TRUE' : 'FALSE'
+      } 
+      AND 
+        "status_audit"."status_id" IN 
+        (${mappedStatusesString}) 
+      AND start_date IS NOT NULL
```
sunghwan-park-fw, line 246: With the next two conditions, this might not need.

apps/api/src/report/report.util.service.ts
```diff
@@ -215,6 +215,120 @@ export class ReportUtilService {
...
+    FROM 
+      "ien_applicant_status_audit" "status_audit" 
+    LEFT JOIN 
+      ien_applicants applicant ON applicant.id = status_audit.applicant_id 
+    WHERE 
+      applicant.new_bccnm_process IS NOT NULL AND applicant.new_bccnm_process = ${
+        BCCNM_NEW_PROCESS ? 'TRUE' : 'FALSE'
+      } 
+      AND 
+        "status_audit"."status_id" IN
```
sunghwan-park-fw, line 244: Instead of passing the map of status IDs, how about this?
```

LEFT JOIN ien_applicant_status ias ON ias.id = status_audit.status_id
WHERE
  applicant.new_bccnm_process = FALSE
  AND
    ias.category = 'IEN Licensing/Registration Process' 
```

apps/api/src/report/report.util.service.ts
```diff
@@ -215,6 +215,120 @@ export class ReportUtilService {
...
+      AND start_date::date >= '${from}' 
+      AND start_date::date <= '${to}' AND 
+      NOT EXISTS (
+        SELECT *
+        FROM "ien_applicant_status_audit" AS T2
+        WHERE 
+          T2.applicant_id = status_audit.applicant_id
+          AND 
+            (
+              T2.start_date > status_audit.start_date
```
sunghwan-park-fw, line 256: This condition means that the milestone should be the latest one. We set an applicant's status to the latest one but don't use it. If we use applicant.status as the latest milestone, couldn't we spare this sub-query? I think we can ignore the cases that 'accept job offer' and 'withdrew from program' are not the latest.

apps/api/src/report/report.util.service.ts
```diff
@@ -215,6 +215,101 @@ export class ReportUtilService {
...
+          BCCNM_NEW_PROCESS ? 'TRUE' : 'FALSE'
+        } AND
+        "status_audit"."status_id" IN ('91F55FAA-C71D-83C8-4F10-3A05E778AFBC','D2656957-EC58-15C9-1E21-3A05E778DC8E' )
+        AND NOT EXISTS (
+          SELECT *
+          FROM 
+            "ien_applicant_status_audit" AS T2
+          WHERE 
+            T2.applicant_id = status_audit.applicant_id
+            AND T2.status_id IN ('f84a4167-a636-4b21-977c-f11aefc486af', '70b1f5f1-1a0d-ef71-42ea-3a0601b46bc2', '632374e6-ca2f-0baa-f994-3a05e77c118a', '18AA32C3-A6A4-4431-8283-89931C141FDE')
```
sunghwan-park-fw, line 269: It's hard to know what these IDs mean. Could we use `STATUS` enum or add comments?

JGH-EY (approved):
Looks good to me.

PR #508, opened 2023-03-13
IEN-775 - replace current applicant id with ATS internal id on table and tombstone

coryatwater-FW:
- add migration to change applicant_id column from uuid to varchar and rename to `ats1_id`
- map ats id to applicant_id 
- make`ats1_id` unique
I kept it nullable, due to the fact theres a lot of old data that has no `ats1_id`, the front end will currently use the `ats1_id` if it's available and if not, use the generated uuid


apps/api/src/applicant/entity/ienapplicant.entity.ts
```diff
@@ -35,7 +35,7 @@ export class IENApplicant {
   name!: string;
 
   @Index({ unique: true })
-  @Column('uuid', { nullable: true })
+  @Column('varchar', { nullable: true })
```
sunghwan-park-fw, line 38: Using the `integer` or `serial` type, we don't need to cast a number to string later.

coryatwater-FW: I normally like to keep things that don't require any mathematics as strings, varchar could be more flexible too.  My opinion, but I can change it too integer if we want, it may require more modifications to other files thinking it's still a `string`

apps/api/src/migration/1678711953274-ChangeApplicantIdColumnToVarChar.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class ChangeApplicantIdColumnToVarChar1678711953274 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "ien_applicants" DROP COLUMN "applicant_id"`);
```
sunghwan-park-fw, line 5: Applicant_id and ats1_id come together from ATS. Considering we take applicant_id as a unique `id,` I wonder if it's necessary to check the uniqueness of `ats1_id`. It seems out of scope to validate the consistency of ATS data.

apps/api/src/applicant/entity/ienapplicant.entity.ts
sunghwan-park-fw: Is there any place referencing this field?
The `applicant_id` field points to `ien_applicants.id` in other tables. To avoid confusion, we could keep using the name `ats1_id`

coryatwater-FW: not in many places, I can change

apps/api/src/migration/1678711953274-ChangeApplicantIdColumnToVarChar.ts
coryatwater-FW: I don't see any harm in keeping it unique as well, under the assumption they're also using it as a unique identifier between ATS and IEN.  @dbayly-freshworks is this correct or at least the intent?

sunghwan-park-fw (approved):

apps/api/src/applicant/external-api.service.ts
```diff
@@ -450,6 +450,7 @@ export class ExternalAPIService {
         assigned_to: { id: string; name: string }[] | undefined;
         registration_date: string;
         applicant_id: string;
+        ats1_id: number;
```
JGH-EY, line 453: Probably missing some context on this one, but you have this listed as string in other uses.

JGH-EY, line 453: Does this need to be number or something? Could this just be a string and then no need to use toString() below?

coryatwater-FW: no they send it to us as a `number`, I typically like to use `strings` for anything that doesn't need any type of computation

PR #509, opened 2023-03-13
IEN-810 Resolve GitHub actions warnings

JGH-EY:
Was able to get the warning resolved for all but one in ZAP. Made the versions all the most recent for the ci-zap file. Found someone mentioning this issue in the GitHub repo for the action. They show that they updated to node 16 and suggest all action runs are displaying this and show where they updated.

### DIFF
--- a/.github/workflows/ci-zap.yml
+++ b/.github/workflows/ci-zap.yml
@@ -3,6 +3,7 @@ on:
   pull_request:
     paths:
       - 'apps/api/**'
+      - 'apps/web/**'
       - 'packages/common/**'
 
 jobs:
--- a/.github/workflows/ci-zap.yml
+++ b/.github/workflows/ci-zap.yml
@@ -17,6 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         name: checkout
+      - uses: actions/setup-node@v3
         with:
           node-version: 16
       - name: ZAP Scan
--- a/.github/workflows/ci-zap.yml
+++ b/.github/workflows/ci-zap.yml
@@ -15,7 +15,7 @@ jobs:
       name: dev
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v2
         name: checkout
 
       - name: ZAP Scan
--- a/.github/workflows/ci-zap.yml
+++ b/.github/workflows/ci-zap.yml
@@ -1,10 +1,10 @@
 name: ZAP Scan
 on:
-  pull_request:
-    paths:
-      - 'apps/api/**'
-      - 'apps/web/**'
-      - 'packages/common/**'
+  push:
+    branches:
+      - main
+    tags:
+      - security
 
 jobs:
   deploy:
@@ -17,9 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         name: checkout
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 16
+
       - name: ZAP Scan
         uses: zaproxy/action-full-scan@v0.4.0
         with:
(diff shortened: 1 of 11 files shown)


apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -81,17 +81,17 @@ describe('Report 4 - Number of IEN registrants in the licensing stage', () => {
     await app.close();
   });
 
+  it('Status 200 on report 4 response', async () => {
+    const { status } = await request(app.getHttpServer()).get(url);
+    expect(status).toBe(200);
+  });
+
   // update table report 4 endpoint reads from
```
JGH-EY, line 89: This was just used to trigger the PR GitHub actions.

apps/api/test/04-report-4.e2e-spec.ts
```diff
@@ -81,17 +81,17 @@ describe('Report 4 - Number of IEN registrants in the licensing stage', () => {
     await app.close();
```
JGH-EY, line 81: This file was only changed to trigger the PR GitHub actions. Changed to original form here

.github/workflows/ci-zap.yml
```diff
@@ -15,11 +15,11 @@ jobs:
       name: dev
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
sunghwan-park-fw, line 18: I wonder if these are required to use a specific node version.
```
- uses: actions/setup-node@v3
  with:
    node-version: 16
```

JGH-EY: Sure, I will test this out.

JGH-EY: It did not work, produced an limit error and still showed the warning about zaproxy/action-full-scan.  To see the warning just look at last failed under ZAP in actions. I think this may be a action-full-scan issue. I found this yesterday, it is recent with the person having the same warning issue.  https://github.com/zaproxy/action-full-scan/issues/61

You can see that they implemented the node upgrade internally. So I don't know if this is fixable at this point.

sunghwan-park-fw: According to the following announcement, I think we don't need to specify github action's local node version if it's 16.
[All Actions will begin running on Node16 instead of Node12](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).

I can't find the warning from the action log, so wonder where it comes.

btw, I don't know why but the action takes too long and fails sometimes.

JGH-EY: Yeah, that's why just adding the latest @v makes it use Node 16.  The maintainers of the different action repos internally set them to use Node 16. That is why there are no warnings on any other action, only the zaproxy/action-full-scan@v0.4.0. Which I mentioned, didn't update node 16 internally this version @0.4.0. 

I don't know what could be done about it taking too long other than increasing the timeout. This is the zaproxy warning, the others would not have warnings.

https://github.com/bcgov/internationally-educated-nurses/actions/runs/4408883243

coryatwater-FW: Hmm, maybe instead of updsting the action versions, we just set `setup-node` to 16 in each workflow, and go back to v2.  There are some new changes in v3 actions that could be hampering performance

sunghwan-park-fw: Let's use `setup-node` only when the action uses nodejs. I found that `setup-node to 16` actually downgrades the version of ubuntu image.

.github/workflows/pr-check-api.yml
```diff
@@ -3,16 +3,21 @@ on:
   pull_request:
     paths:
       - 'apps/api/**'
+      - 'apps/web/**'
```
coryatwater-FW, line 6: don't need this line

PR #511, opened 2023-03-13
Seed test data

sunghwan-park-fw:
- Create applicants
- Create milestones to distribute applicants evenly over each stage.
- Randomly skip milestones

### DIFF
--- a/packages/scripts/src/generate-milestones.ts
+++ b/packages/scripts/src/generate-milestones.ts
@@ -7,39 +7,37 @@ import _ from 'lodash';
 import {
   ApplicantJobRO,
   ApplicantRO,
-  BCCNM_NCAS_STAGE,
   IENApplicantAddStatusDTO,
+  IENApplicantCreateUpdateDTO,
   IENApplicantJobCreateUpdateDTO,
   IENHaPcnRO,
   IMMIGRATION_STAGE,
   LIC_REG_STAGE,
-  NNAS_STAGE,
   RECRUITMENT_STAGE,
   STATUS,
 } from '@ien/common';
 
 config({ path: '../../.env' });
+
+const MIN_DURATION = 5;
+const MAX_DURATION = 20;
+const APPLICANT_COUNT = 4000;
 let start = new Date();
 
-const AUTH_URL = 'https://keycloak.freshworks.club/auth';
-
 const getToken = async (): Promise<void> => {
-  const url = `${AUTH_URL}/realms/ien/protocol/openid-connect/token`;
-  const resp = await axios.post(
-    url,
-    {
-      grant_type: 'password',
-      client_id: 'IEN',
-      username: process.env.E2E_TEST_USERNAME,
-      password: process.env.E2E_TEST_PASSWORD,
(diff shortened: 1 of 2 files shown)


packages/scripts/src/generate-milestones.ts
```diff
@@ -148,23 +165,73 @@ const checkToken = async (start: Date) => {
...
+    first_name,
+    last_name,
+    email_address: faker.internet.email(first_name, last_name),
+    phone_number: faker.phone.number('###-###-####'),
+    registration_date: dayjs(registration).format('YYYY-MM-DD'),
+    nursing_educations: [],
+    country_of_citizenship: [faker.helpers.arrayElement(Object.keys(COUNTRY_OF_EDUCATIONS))],
+    country_of_residence: faker.helpers.arrayElement(Object.keys(COUNTRY_OF_EDUCATIONS)),
+    pr_status: 'No status',
+    is_open: true,
```
coryatwater-FW, line 197: we probably want the `new_bccnm_process` column too

packages/scripts/src/generate-milestones.ts
```diff
@@ -148,23 +165,73 @@ const checkToken = async (start: Date) => {
...
-      const applicant = applicants[i];
-      if (dayjs(applicant.registration_date).isAfter('2022-06-01')) {
-        console.log(`skip applicant registered at ${applicant.registration_date}`);
-        continue;
-      }
+      const applicant = applicantsWithoutMilestones[i];
 
-      await generateMilestones(applicant.id, statuses);
-      console.log(`${i}/${count} applicant processed <- ${applicant.id}`);
+      const numOfMilestones = await generateMilestones(applicant.id, statuses);
```
coryatwater-FW, line 233: `addApplicant` returns an `ApplicantRO` couldn't we just pass the applicant instead of the id and having to use `await getApplicant(id)` in `generateMilestones`?

sunghwan-park-fw: The applicants list contains the result of `getApplicants` that do not have milestones.

coryatwater-FW: yeah, that's what I was asking.  Was curious if `applicantsWithoutMilestones` is just a filtered list of `ApplicantRO` then we could just pass the applicant object instead of the id? As the `generateMilestones` fn is calling `getApplicant(id)`, just wondering if it's returning data we already had before calling

sunghwan-park-fw: You're right if we generate milestones only for the newly added applicants.

packages/scripts/src/generate-milestones.ts
```diff
@@ -148,23 +165,73 @@ const checkToken = async (start: Date) => {
...
-    const [applicants, count] = await getApplicants();
+    const [applicants, initialCount] = await getApplicants();
+
+    console.log(`${initialCount} applicants found`);
+
+    while (applicants.length < APPLICANT_COUNT) {
+      const applicant = await addApplicant('2021-01-01', '2022-01-01');
+      applicants.push(applicant);
+    }
+    const count = applicants.length;
```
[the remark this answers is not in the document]
sunghwan-park-fw: Two points are using the count.

PR #514, opened 2023-03-14
IEN-811 Updated class-validator in dependencies

JGH-EY:
Updated to dependabots suggested version for class-validator


sunghwan-park-fw (approved):

sunghwan-park-fw (approved):
It looks good to me.

PR #532, opened 2023-04-11
Ien 781 data mismatch in reports

dbayly-freshworks:
I replaced the query for report 5 with something a bit simpler and faster. 

Added columns for BCCNM new process for report 5.

### DIFF
--- a/apps/api/src/report/report.service.ts
+++ b/apps/api/src/report/report.service.ts
@@ -401,6 +401,7 @@ export class ReportService {
    * @returns
    */
   async getLicenseApplicants(f: string, t: string) {
+    const { from, to } = this.captureFromTo(f, t);
     const statuses = await this.getStatusMap();
     const licence_statuses = [
       statuses[STATUS.BCCNM_FULL_LICENSE_RN],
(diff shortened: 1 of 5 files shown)


apps/api/src/report/report.service.ts
```diff
@@ -401,23 +401,49 @@ export class ReportService {
    * @returns
    */
   async getLicenseApplicants(f: string, t: string) {
-    const { from, to } = this.captureFromTo(f, t);
```
coryatwater-FW, line 404: wouldn't this be good incase `f` and `t` are undefined?

apps/api/src/report/report.util.service.ts
```diff
@@ -325,6 +325,45 @@ export class ReportUtilService {
...
+        FROM 
+          "ien_applicant_status_audit" "status_audit"
+        LEFT JOIN 
+          ien_applicants applicant ON applicant.id = status_audit.applicant_id  
+        WHERE 
+          applicant.new_bccnm_process IS NOT NULL 
+          AND applicant.new_bccnm_process = ${BCCNM_NEW_PROCESS} 
+          AND "status_audit"."status_id" 
+          IN (
+            \'${statuses.join("','")}\'
```
coryatwater-FW, line 357: does this ened to be an array? just looks like we're passing a single value

dbayly-freshworks: Yup. I'll update it. Originally, I intended this function to be able to handle multiple statues, (eg do all full licences at the same time)

PR #536, opened 2023-05-01
Ien 840 add milestone extract

dbayly-freshworks:


apps/api/src/report/report.service.ts
```diff
@@ -765,6 +770,20 @@ export class ReportService {
     return data;
   }
 
+  /**
+   *
+   * @param dates start and end date for data extract YYYY-MM-DD
+   * @returns
+   */
+  async extractMilestoneData(dates?: ReportPeriodDTO) {
+    const { from, to } = dates || {from:'2001-01-01', to:dayjs().toISOString()};
```
sunghwan-park-fw, line 779: It might not make any difference, but to match the pattern, how about `dayjs().format('YYYY-MM-DD')`?

apps/api/src/report/report.service.ts
```diff
@@ -737,6 +737,11 @@ export class ReportService {
     return data;
   }
 
+  /**
+   *
+   * @param dates start and end date for data extract YYYY-MM-DD
```
sunghwan-park-fw, line 742: Duplicate comments?

sunghwan-park-fw (approved):
Just minor comments.

PR #545, opened 2023-05-25
IEN-842 list user guide versions

sunghwan-park-fw:
- Add API endpoints
  - to get versions of a file
  - to delete a version or a file
  - to restore a file to a version
- Add `version` option to the download endpoint 
- Use `Disclosure` to display versions

### DIFF
--- a/apps/web/src/components/admin/UserGuideRow.tsx
+++ b/apps/web/src/components/admin/UserGuideRow.tsx
@@ -1,28 +1,146 @@
 import dayjs from 'dayjs';
+import { useEffect, useState } from 'react';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faReply, faSpinner, faTrash } from '@fortawesome/free-solid-svg-icons';
+import { mutate } from 'swr';
 import { UserGuide } from '@ien/common';
-import { downloadGuide } from '../../services/admin';
+import { Disclosure } from '@components';
+import {
+  deleteUserGuide,
+  downloadUserGuide,
+  getUserGuideVersions,
+  restoreUserGuide,
+} from '../../services/admin';
 
 interface UserGuideProps {
   file: UserGuide;
+  showVersions: boolean;
 }
 
-export const UserGuideRow = ({ file: { name, lastModified } }: UserGuideProps) => {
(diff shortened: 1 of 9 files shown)


apps/web/src/components/admin/UserGuideRow.tsx
```diff
@@ -1,28 +1,146 @@
...
-  const open = async () => {
-    const url = await downloadGuide(name);
-    window.open(url);
+export const UserGuideRow = ({ file, showVersions }: UserGuideProps) => {
+  const [versions, setVersions] = useState<UserGuide[]>();
+  const [loading, setLoading] = useState(false);
+  const [selectedVersion, setSelectedVersion] = useState<string | undefined>();
+
+  const getVersions = (open: boolean) => {
+    if (open) {
```
coryatwater-FW, line 26: whats the purpose of the `boolean`?

apps/web/src/components/admin/UserGuideRow.tsx
```diff
@@ -1,28 +1,146 @@
...
+  };
+
+  useEffect(() => {
+    if (!loading) setSelectedVersion(undefined);
+  }, [loading]);
+
+  const getUserGuideRow = ({ name, lastModified, version, size }: UserGuide, index = -1) => {
+    const padding = version ? 'py-2 pl-3' : '';
+    return (
+      <div className={`flex flex-row align-middle ${padding}`} key={name + version}>
```
coryatwater-FW, line 83: are names unique? wonder if this could throw non-sense console errors complaining about the key

apps/web/src/components/admin/UserGuideRow.tsx
sunghwan-park-fw: If the disclosure is expanded, then `open` is true and we need to query versions to display them.

apps/web/src/components/admin/UserGuideRow.tsx
sunghwan-park-fw: I believe the name(aws object key) should be unique. If we upload one with the same name, then the old one becomes the old version of the new file.

coryatwater-FW (approved):

PR #551, opened 2023-06-15
IEN-851 add ien type field to milestone

sunghwan-park-fw:
- Add IEN type field to `ien_applicant_status_audit` table
- Add a radio input to the milestone form
- Fix validation error of creating and updating milestone

Start and effective dates are mandatory for Job Offer Accepted but it hasn’t been validated correctly. It’s because the validator has the conditions comparing milestone name instead of id, but we’ve been sending status id in the form data.

Now we send status name to the backend.


dbayly-freshworks (approved):
Looks ok to me!

PR #553, opened 2023-06-21
IEN-852 add IEN type to data-extract

sunghwan-park-fw:


apps/api/src/report/report.service.ts
```diff
@@ -764,6 +785,15 @@ export class ReportService {
     const data = await entityManager.query(
       this.reportUtilService.extractApplicantsDataQuery(from, to, milestones),
     );
+
+    // set IEN type
+    const types = await this.getIenTypes(from, to);
```
dbayly-freshworks, line 790: Why do we need a second call to fetch the types data? Shouldn't the types information be included in `data` variable?

sunghwan-park-fw: I couldn't find a proper way to do that. The query to get the `data` is already a little heavy, taking about 8 seconds, and I didn't want to make it more complex and heavy unless it is possible just by a select field, as I did for the milestone report.

dbayly-freshworks (approved):

PR #558, opened 2023-07-05
IEN-844 - preliminary hide inactive and active/inactive applicant toggle

coryatwater-FW:
- added checkbox to hide Inactive applicants from manage applicants table
- added toggle on applicant details page to switch active flag between active and inactive
- add `is_active` flag to applicant table

- this is just preliminary work, with the assumption of future discussions to add logic and determine what makes an applicant `inactive`

[IEN-844](https://eydscanada.atlassian.net/jira/software/c/projects/IEN/boards/272?modal=detail&selectedIssue=IEN-844)


dbayly-freshworks (approved):
I pulled this down and it works well!  I'll pass it by Shawn/the PO and see what they think of the workflow/design.

PR #559, opened 2023-07-05
IEN-847 Add ability to assign applicant to me

sunghwan-park-fw:
- add ien_applicants_recruiters_employee table relating applicants to ha users
- delete recruiter_name from job (recruiter is not from job-level but from HA level)
- add endpoint /ien/:id/assign to assign an applicant to the current login user with APPLICANT_WRITE access
- add migration to set HA user’s organization if it is null
- upgrade cypress from 9 to 12


apps/api/src/applicant/ienapplicant.service.ts
```diff
@@ -535,4 +538,32 @@ export class IENApplicantService {
...
+
+  async assignApplicant(id: string, employee: EmployeeRO) {
+    const applicant = await this.ienapplicantRepository.findOne(id);
+
+    const found = applicant?.recruiters?.some(e => e.id === employee.id);
+    if (found) return;
+
+    const haPcn = await this.haPcnRepository.findOne({ title: employee.organization });
+    if (!haPcn) {
+      throw new BadRequestException(`User doesn't belong to a health authority`);
```
coryatwater-FW, line 550: small nit - param is called employee but we're throwing an error calling it a user, also maybe add a few comments just to showcase what's happening in this fn

apps/api/src/applicant/ienapplicant.service.ts
```diff
@@ -535,4 +538,32 @@ export class IENApplicantService {
...
+    if (!recruiterAssignment) {
+      recruiterAssignment = await this.recruiterRepository.create({
+        applicant_id: id,
+        ha_id: haPcn.id,
+        employee_id: employee.id,
+      });
+    } else {
+      recruiterAssignment.employee_id = employee.id;
+    }
+    return await this.recruiterRepository.save(recruiterAssignment);
```
coryatwater-FW, line 567: should we wrap this in a transaction incase it fails?

apps/api/src/migration/1688145070120-RelateIenToHA.ts
```diff
@@ -0,0 +1,26 @@
...
+        applicant_id uuid,
+        ha_id uuid,
+        employee_id uuid,
+        PRIMARY KEY("applicant_id", "ha_id"),
+        FOREIGN KEY("applicant_id") REFERENCES "ien_applicants"("id") ON DELETE CASCADE,
+        FOREIGN KEY("ha_id") REFERENCES "ien_ha_pcn"("id") ON DELETE CASCADE,
+        FOREIGN KEY("employee_id") REFERENCES "employee"("id") ON DELETE CASCADE
+      );
+      DROP TABLE IF EXISTS "ien_applicants_assignments_employee";
+      DROP TABLE IF EXISTS "ien_applicants_assignments_employee";
```
coryatwater-FW, line 17: got same query twice here

apps/api/src/applicant/ienapplicant.service.ts
sunghwan-park-fw: emplyee.organization column is not related to haPcn by a foreign key constraint, so it could be any string.

apps/api/src/applicant/ienapplicant.service.ts
sunghwan-park-fw: I don't think we need a transaction for a single write query.

apps/web/src/components/applicant/ApplicantProfile.tsx
```diff
@@ -84,6 +75,31 @@ export const ApplicantProfile = () => {
                 .join(', ')
             }
           />
+          <DetailsItem title='Assigned To'>
+            <div>
+              <span className='border border-gray-200 bg-bcGrayLabel text-white rounded text-xs px-2 py-0.5 mr-1'>
+                HMBC
```
coryatwater-FW, line 81: curiosity questions, is showing HMBC necessary once they're assigned to a HA? and should HMBC/ MoH be able to see who they're assigned too, maybe not show the name, but could be an easier way to view what HA's the applicant is apart of, especially since there's no search for licensing/ registration?

coryatwater-FW: no no, I meant from a HMBC perspective, for example if I log in as a HA assign a user to me, you can see `HMBC` and the HA the user is from, then when you sign back in as HMBC you only see HMBC.  Just a thought for more clarity for HMBC, not a blocker though.

sunghwan-park-fw: It's the same for HMBC and HAs. Recruiter is meant to be used only for HAs, so I don't think we need to let HMBC know each HA's recruiters.

PR #562, opened 2023-07-06
IEN-844 - move active flag to own table, fix HA level

coryatwater-FW:
- moved `active_flag` from applicants table to it's own table to allow for HA level inactive/ active statuses


dbayly-freshworks (approved):
This looks ok to me

PR #567, opened 2023-07-13
IEN-853 add inferred type column to extract-applicant report

sunghwan-park-fw:
- Add Inferred Type column next to `Type` column in the applicant-data-extract spreadsheet
- If `job offer accepted milestone` doesn't have `stream-type`, it is decided by the latest licensing milestone.


jerry-ey:
close as it seems outdated

PR #576, opened 2023-08-14
Ien 860 milestone update

dbayly-freshworks:
* Added a version column to separate new and old milestones
* Marked old milestones as version 1
* Added new milestones for version 2 
* Updated export to only include milestones from version 2 
* Updated FE constants 
* Updated name of withdrawal status to "Not proceeding"


apps/api/src/migration/1663698520477-RemoveDuplicateMilestones.ts
```diff
@@ -34,17 +34,22 @@ export class RemoveDuplicateMilestones1663698520477 implements MigrationInterfac
...
-        category: 'IEN Recruitment Process',
-      })
-      .execute();
+    try {
+      await queryRunner.manager
+        .createQueryBuilder()
+        .insert()
+        .into('ien_applicant_status')
+        .values({
+          id: '3fd4f2b0-5151-d7c8-6bbc-3a0601b5e1b0',
```
sunghwan-park-fw, line 43: Why do we need to insert this? I see it on my local database.

apps/api/src/migration/1692040928087-RenameWithdrawStatus.ts
```diff
@@ -0,0 +1,17 @@
+import { IENApplicantStatus } from 'src/applicant/entity/ienapplicant-status.entity';
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class RenameWithdrawStatus1692040928087 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.connection
+      .createQueryBuilder()
+      .update(IENApplicantStatus)
+      .set({
+        status: 'Not Proceeding',
```
sunghwan-park-fw, line 10: Don't we need to change `STATUS.WITHDREW_FROM_PROGRAM` too? It is also referenced in many places.

dbayly-freshworks: I changed the value of WITHDREW_FROM_IEN_PROGRAM, but not the key. I have mixed feelings about doing that as it's a very large change.

apps/api/src/migration/1663698520477-RemoveDuplicateMilestones.ts
dbayly-freshworks: I had to add this because this migration was breaking in the pipeline. I think the query builder referenced the entity when constructing the query. By adding the version column to the entity, this migration tried to add a version to the insert query. I patched it by using a direct query instead. 
Screenshot 2023-08-14 at 3 51 45 PM

apps/api/src/migration/1692040928087-RenameWithdrawStatus.ts
sunghwan-park-fw: I got it, but at least doesn't STATUS need to be updated like this?
`WITHDREW_FROM_PROGRAM = 'Not Proceeding'`

sunghwan-park-fw (approved):

PR #579, opened 2023-08-25
IEN-855 Add BCCNM/NCAS role

sunghwan-park-fw:
- Add BCCNM/NCAS role with BCCNM/NCAS access
- Make BCCNM/NCAS role dependent on `Manage Applicants` role
- Fix the loading state of user guides
- Deployed to dev


dbayly-freshworks (commented):
We need to be careful with this role. We should restrict the permisson to assign this role only to super-admins. 
Otherwise, I tested this out on dev and it works well.

dbayly-freshworks (approved):

sunghwan-park-fw:
> We need to be careful with this role. We should restrict the permisson to assign this role only to super-admins. Otherwise, I tested this out on dev and it works well.

Updated `user details` page so that BCCNM/NCAS role could be displayed only for admin users.

PR #580, opened 2023-08-29
IEN-867 fix applicants query

sunghwan-park-fw:
- Parameterize query to avoid SQL injection


dbayly-freshworks (approved):
Looks good to me!

PR #590, opened 2023-09-14
IEN-873 feedback bccnm ncas updates

sunghwan-park-fw:


dbayly-freshworks (approved):
This is a big PR, but it looks ok to me.

PR #591, opened 2023-09-20
IEN-875 replace 'referred to ncas' with 'applied to ncas'

sunghwan-park-fw:


apps/api/src/migration/1676405526677-CreateReportView.ts
```diff
@@ -1,444 +0,0 @@
-import _ from 'lodash';
-import { MigrationInterface, QueryRunner } from 'typeorm';
-import { STATUS } from '@ien/common';
```
dbayly-freshworks, line 3: Whats the purpose of removing this migration?

sunghwan-park-fw: I thought we'd better remove STATUS.REFERRED_TO_NCAS to avoid confusion and this migration references it a lot. So I thought it's better to remove it cause it doesn't affect the sequence of migrations.

dbayly-freshworks (approved):
Looks good overall aside from the deleted migration.

PR #594, opened 2023-09-22
IEN-864 soft-delete and history of milestones

sunghwan-park-fw:
- Add deleted_date column to `ien_applicant_status_audit` table
- Add `milestone_audit` table to store history of milestone changes
- Add a subscriber to insert entries to `milestone_audit` when a milestone is updated
- When a milestone is created, if it conflicts with one deleted, it is restored and updated with the new one.


jerry-ey:
close as it seems outdated

PR #603, opened 2023-11-09
IEN-884 remove duplicate status and add foreign key for status_id

sunghwan-park-fw:


dbayly-freshworks (approved):
This all looks good to me

PR #606, opened 2023-12-12
IEN patch add generic education fix import

dbayly-freshworks:
Fixed an import issue
Changed the label for educations from the BCCNM

### DIFF
--- a/apps/api/src/applicant/ienapplicant.service.ts
+++ b/apps/api/src/applicant/ienapplicant.service.ts
@@ -618,7 +618,7 @@ export class IENApplicantService {
     }
 
     nursing_educations.push({
-      name: 'Diploma/Certificate of Nursing',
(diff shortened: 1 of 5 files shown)


apps/api/src/applicant/ienapplicant.service.ts
```diff
@@ -618,7 +618,7 @@ export class IENApplicantService {
     }
 
     nursing_educations.push({
-      name: 'Diploma/Certificate of Nursing',
+      name: 'Education',
```
sunghwan-park-fw, line 621: Does this mean the case that they don't know the title of education?

dbayly-freshworks: That is correct.

sunghwan-park-fw (approved):

PR #612, opened 2024-01-22
Ien 886 fix sync

dbayly-freshworks:
Fixes an issue where the prod sync was failing due data from ATS violating the unique key constraint on ats1_id.
Added a function to filter out rows that cause a contradiction with existing data. 

Commented out sync due issues with dev/test sync. Syncs will be run manually for the time being.


apps/api/src/applicant/external-api.service.ts
```diff
@@ -512,7 +514,33 @@ export class ExternalAPIService {
...
+   * @returns A filtered list of applicants that will not cause a violation of the unique key constraint for the ats1_id
+   */
+  async filterContradictoryRows(applicants: AtsApplicant[]) {
+    const existingApplicants: IENApplicant[] = await this.ienapplicantRepository.find({});
+    const contradictions: unknown[] = [];
+    const filteredApplicants = applicants.filter(applicant => {
+      const contradiction = existingApplicants.find(
+        existingApplicant =>
+          existingApplicant.ats1_id === applicant.ats1_id.toString() &&
+          existingApplicant.id !== applicant.applicant_id,
```
coryatwater-FW, line 530: this is just filtering one of the existing applicants out? aren't they two different applicants or is this referring to same applicant with the same `ats1_id`?

dbayly-freshworks: This is for finding an existing applicant with the same ats1_id but different applicant id in the system. 
On prod there is at least one row that has this issue, and it's preventing the sync from completing.

coryatwater-FW (approved):

PR #621, opened 2024-07-10
IEN-903 | add new communities in HA NorthHealth

jerry-ey:
Ticket: https://eydscanada.atlassian.net/browse/IEN-903

## Changes
- Add new communities in North Health

## Notes
- I have tested locally with this migration, also tested migration:revert
- For the community not listed in ticket IEN-903, I discussed with Ram that the existing community will remain in place to avoid any negative impact. That's why I only have "insert" script.

- Screenshot of local database

CleanShot 2024-07-10 at 11 31 05

CleanShot 2024-07-10 at 11 35 54


dbayly-freshworks (approved):
This looks good. The only thing that worries me is the hard coded ids. There is a chance that the migration might fail if the target environment has a community with a matching id. Lets keep an eye out for this when deploying.

PR #628, opened 2024-07-24
IEN Terraform updates 

dbayly-freshworks:
Fixes: 
Updated all lambda function runtimes to be nodejs18x. 
Updated aws terraform version to version 4. 
Updated RDS database sizes to be smaller, t3 mediums instead of r5 larges.


jerry-ey (approved):
Looks great!

PR #634, opened 2024-08-09
IEN-890 | Add 3 new columns for BCCNM spreadsheet upload

jerry-ey:
## Requirements
- Added 3 new statuses.

## Enhancements
- Improved support for large spreadsheets: [Stream large data with SheetJS](https://docs.sheetjs.com/docs/demos/bigdata/stream/).

## Bugs Fixed (not on the ticket)
While developing, I encountered two bugs:
- The date in the spreadsheet shows a `1-day` difference from the Preview Table.
- Duplicate `HMBC Unique ID` values in the spreadsheet cause table key errors, leading to rendering issues when navigating between pages.

## Github Action fixed
- It turns out not the e2e testing failed, but somehow `docker-compose` is not being used, it has to be `docker compose` in `ubuntu-latest`
https://github.com/orgs/community/discussions/116610


jerry-ey:
@dbayly-freshworks 

I have deployed this PR to TEST and upload a file. Also connected to TEST DB to change myself to become admin user.
It's working great.

CleanShot 2024-08-13 at 13 58 35

apps/web/src/components/admin/UserGuideList.tsx
dbayly-freshworks, line 1: Looks like this file is empty, we can probably delete this.

apps/api/package.json
```diff
@@ -55,6 +55,7 @@
     "swagger-ui-express": "4.2.0",
     "typeorm": "0.2.41",
     "winston": "3.3.3",
+    "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz",
```
dbayly-freshworks, line 58: Any reason why we're pulling xlsx from a cdn instead from an npm repository?

dbayly-freshworks (approved):
Two comments, otherwise this looks good to deploy.

apps/api/package.json
jerry-ey: yeah, I had the same feeling when I saw this at firs time. It's actually from the official doc.
https://docs.sheetjs.com/docs/getting-started/installation/nodejs

CleanShot 2024-08-13 at 14 53 41

dbayly-freshworks: Oh, thats very strange. 

All good though

apps/web/src/components/admin/UserGuideList.tsx
jerry-ey: fixed! somehow when I restore that file from main branch, it changed the file permission. (The reason I made change locally is because I have to stop the userGuide api call from the frontend locally, otherwise it kept showing error message because of s3 connection and block my development)
This was the change, and I accidentally commit and push to this feature branch, then I restore that file from main branch : 
CleanShot 2024-08-13 at 15 08 38


CleanShot 2024-08-13 at 15 05 34


The change from `100644` to `100755` indicates a modification in the file permissions, specifically making the file executable. In Unix-based systems, `100644` represents a regular file that is not executable, while `100755` means the file is executable by the owner, group, and others.

### Why Did This Happen?

When you restored the file from the main branch, the content was restored, but the file permissions might not have reverted automatically. This can happen if the permissions were changed locally before the restore and then committed.

### Steps to Fix This in Your GitHub PR

To revert the file permissions back to `100644` (non-executable), follow these steps:

1. **Change the File Permissions Locally:**
   
   Run the following command in your terminal to change the permissions back to `644`:

   ```bash
   chmod 644 apps/web/src/components/admin/UserGuideList.tsx
   ```

2. **Check the File Status:**

   After changing the permissions, check the status of the file with:

   ```bash
   git status
   ```

   You should see that Git recognizes the change in file permissions.

3. **Commit the Changes:**

   Add the file and commit the change:

   ```bash
   git add apps/web/src/components/admin/UserGuideList.tsx
   git commit -m "Revert file permissions for UserGuideList.tsx"
   ```

4. **Push the Changes:**

   Push the commit to the branch that is associated with the PR:

   ```bash
   git push
   ```

### Verify the Changes on GitHub

Once pushed, go back to your GitHub PR and verify that the file no longer shows the `100644 → 100755` change. The PR should now only reflect the intended content changes (if any) without altering the file permissions.

This should resolve the issue and ensure that the file permissions are consistent with the main branch.

dbayly-freshworks: Oh thats interesting, I've seen this in a few prs lately, and I've been wondering how it happens.

jerry-ey: yeah, my first time. glad chatgpt can answer this :)

PR #638, opened 2024-08-19
Added HA specific extracts for both milestone and applicant extracts

dbayly-freshworks:
Updated data extracts. For HA users, only users who have a "referred to " milestone will have their milestones extracted. 

I added two steps to the process. The first takes the users HA PCN and finds the id of the "Applicant Referred to " milestone for that HA. The second selects a list of ids of users who have that milestone. 

The remainder is the main query, just with a check to ensure that only milestones from those users are selected.

### DIFF
--- a/apps/web/src/pages/_app.tsx
+++ b/apps/web/src/pages/_app.tsx
@@ -23,7 +23,7 @@ function App({ Component, pageProps }: AppProps) {
 
   useEffect(() => {
     setOidcConfig({
-      authority: `https://common-logon-dev.hlth.gov.bc.ca/auth/realms/moh_applications`,
+      authority: `${process.env.NEXT_PUBLIC_AUTH_URL}/realms/${process.env.NEXT_PUBLIC_AUTH_REALM}`,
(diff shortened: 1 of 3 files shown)


jerry-ey:
Linting issue
CleanShot 2024-08-19 at 14 12 03@2x

jerry-ey (approved):
Looks great to me! And thanks for the help :)

apps/web/src/pages/_app.tsx
```diff
@@ -23,7 +23,7 @@ function App({ Component, pageProps }: AppProps) {
 
   useEffect(() => {
     setOidcConfig({
-      authority: `${process.env.NEXT_PUBLIC_AUTH_URL}/realms/${process.env.NEXT_PUBLIC_AUTH_REALM}`,
+      authority: `https://common-logon-dev.hlth.gov.bc.ca/auth/realms/moh_applications`,
```
jerry-ey, line 26: you might accidentally commit this one

dbayly-freshworks: Thank you! I will remove it

PR #639, opened 2024-08-20
Ien sync prototype fix

dbayly-freshworks:
Re-added sync. It looks like the data in the sync has been restored. 
Added a function to run the sync as a smaller syncs instead of a single 
I have wrapped this feature in a flag, so it will be disabled by default.

### DIFF
--- a/apps/api/src/syncdata.ts
+++ b/apps/api/src/syncdata.ts
@@ -55,7 +55,6 @@ export const handler: Handler = async (event, context: Context) => {
                 result = await externalAPIService.saveApplicant(from, to, page);
               } catch (e) {
                 logger.error(e, `ATS-SYNC Page ${page} failed.`);
-                console.log(`Page ${page} failed.`, e);
                 failCount++;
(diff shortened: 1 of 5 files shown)


apps/api/src/syncdata.ts
```diff
@@ -44,7 +45,26 @@ export const handler: Handler = async (event, context: Context) => {
...
+          let result: SyncApplicantsResultDTO | undefined;
+          if (!page) {
+            page = 1;
+            let failCount = 0;
+            do {
+              try {
+                result = await externalAPIService.saveApplicant(from, to, page);
+              } catch (e) {
+                logger.error(e, `ATS-SYNC Page ${page} failed.`);
+                console.log(`Page ${page} failed.`, e);
```
jerry-ey, line 58: maybe add 
`// eslint-disable-next-line no-console` to avoid lint issue

jerry-ey (approved):
Looks good to me :)

PR #643, opened 2024-08-21
Made country of education optional, made tweak to sync

dbayly-freshworks:
Having a row without a country of education was blocking the sheet upload. 
I also needed to make some small tweaks to the sync.


jerry-ey (approved):
Looks good!

PR #645, opened 2024-08-22
IEN-897 | Disable the period report cron job

jerry-ey:
- It turns out comment out the `schedule_expression` failed deployment.
- So we add `count = 0` instead to disable it. And by using `count`, we have to add `[count.index]`.

I have verified by deploying this PR to DEV and see from the AWS console.

### AWS DEV (After deployed this PR)

CleanShot 2024-08-22 at 10 38 42

### AWS TEST (Before this PR)

CleanShot 2024-08-22 at 10 40 25


dbayly-freshworks (approved):
Looks good to me!

PR #646, opened 2024-08-23
IEN-885 | Fix Report Large Response issue

jerry-ey:
- Add new S3 bucket to save the large json
- Lambda upload json to S3 and return pre-signedUrl to frontend to download
- Only when data length is more than 10 and not in local and test env would use s3 to download the file

I did deploy to TEST and works fine.

### New bucket
CleanShot 2024-08-23 at 15 55 43

### Large spreadsheet
- it has 22133 rows

CleanShot 2024-08-23 at 15 51 56

### DIFF
--- a/apps/api/src/report/report.s3.service.ts
+++ b/apps/api/src/report/report.s3.service.ts
@@ -25,8 +25,6 @@ export class ReportS3Service {
         Bucket: BUCKET_NAME,
         Key: key,
         Body: JSON.stringify(data),
-        ContentType: 'application/json',
-        ACL: 'public-read',
       };
       await this.s3.upload(params).promise();
     } catch (e) {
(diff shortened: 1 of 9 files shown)


apps/api/src/report/report.s3.service.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  constructor() {
+    this.s3 = new AWS.S3({
+      params: {
+        Bucket: BUCKET_NAME,
+      },
+    });
+  }
+
+  async uploadFile(key: string, data: any): Promise<void> {
```
dbayly-freshworks, line 19: If we can provide a type or even just some of the params, that would be great.

dbayly-freshworks (approved):
This looks mostly good. 
Two small comments. 
Can we look into ways to automate deleting reports for the bucket?

jerry-ey:
Sounds good. Thanks for checking.

Will look into automatic delete and address 2 comments.

jerry-ey:
## New addition
- Add s3 life cycle: all objects will be deleted after 1 day
- Fix data any type
- Fix s3 origins with frontend url

PR #648, opened 2024-09-03
Fixed off by one error found in sync

dbayly-freshworks:
Fixed an issue where the last user in the sync would not be upserted.


jerry-ey (approved):
Looks good!

PR #654, opened 2024-09-18
IEN-885 | large response fix

jerry-ey:
## Issue

- The Lambda function responsible for uploading files to S3 is taking too long, causing timeout errors.
- Even after increasing the Lambda and CloudFront response timeouts, the issue persists.
- Large files take at least 2 minutes to upload, but CloudFront has a maximum response timeout of **60 seconds**.

## Approach

- Created a second Lambda function to handle the file upload in the background.
- The API Lambda invoke the Upload Report Lambda, and then returns a pre-signed URL first.
- The frontend polls the pre-signed URL until the file upload is complete.
- Polling is used as we do not have WebSocket support.

## Result 

Have deployed to TEST and working fine. Beware that for downloading applicant would take you around 2 mins.

CleanShot 2024-09-18 at 15 15 02

---

New lambda:

CleanShot 2024-09-18 at 15 18 08


apps/api/src/report/report.controller.ts
```diff
@@ -171,16 +171,65 @@ export class ReportController {
...
+  }
+
+  private shouldUseS3(): boolean {
+    return process.env.NODE_ENV !== 'test' && process.env.RUNTIME_ENV !== 'local';
+  }
+
+  private generateS3Key(
+    from: string,
+    to: string,
+    userId: string | undefined | null,
```
dbayly-freshworks, line 190: Nit - It seems strange that user Id could be both undefined or null. 
If the user id is undefined, shouldn't we throw an error?

dbayly-freshworks (approved):
Overall looks good, just one comment.

apps/api/src/report/report.controller.ts
jerry-ey: Thanks, good catch! fixed.

PR #657, opened 2024-09-25
IEN-914 | Remove communities from Northern Health Authority that are not on the list.

jerry-ey:
- Have tested locally and double check with the original list. (https://eydscanada.atlassian.net/browse/IEN-903)

CleanShot 2024-09-25 at 11 38 24


dbayly-freshworks (approved):
This looks ok to me. 
What is the plan for any records who have been assigned one of the invalid community labels?

jerry-ey:
@dbayly-freshworks 
Thanks David, I have checked the schema. Also tested this locally, the community will be N/A showing in frontend.

There is `on delete cascade`, so once the community is removed, then the related data will be deleted too. So we don't need to worry for any records who have been assigned one of the invalid community labels.
CleanShot 2024-09-26 at 10 36 33

PR #659, opened 2024-09-27
IEN-921 | add limit as required

jerry-ey:
- As discussed, the endpoint functions correctly. We’ve added limit as a required field to prevent failures caused by large data responses.
- Since class-validator does not support overriding decorators, we had to create a base class instead of directly inheriting it.
(https://github.com/typestack/class-validator/issues/156)


dbayly-freshworks (approved):
Looks ok to me1

PR #656, opened 2024-09-24
IEN-923 | Enable HA Report Permission

jerry-ey:
## Target
- Add back the report permission to HA user.
- It's most of the revert of this past PR (https://github.com/bcgov/internationally-educated-nurses/pull/592/files)

## Query
- The query inserts all HA employees without "reporting" role.

## Screenshots

- Have tested locally
CleanShot 2024-09-24 at 11 14 59

- The report already only download the associated organization

CleanShot 2024-09-24 at 11 17 37

(Login myself as Fraser Health Authority)
CleanShot 2024-09-24 at 11 19 51

PR #661, opened 2024-10-10
IEN-928 | Add Feature Flag to HA access Report

jerry-ey:
After digging more, I decide to use `parameter store` with `dynamic fetch the parameter` in api so that in the future we don't encounter terraform state inconsistent and easier to add more other feature flags.

- Because it's hard to add feature flag to the migration(previous enable HA PR: https://github.com/bcgov/internationally-educated-nurses/pull/656), so we implement this in api level to filter user role if it's HA user and with reporting role.
- If in local env, ignore the parameter store and always return true
- Have deployed to DEV and verified.
  - Tested redeploy terraform not override the existing value
  - Tested the feature flag functionality by toggling the value in parameter store

## Screenshots

- lambda can read the parameter

CleanShot 2024-10-10 at 13 22 48

- parameter store

CleanShot 2024-10-10 at 13 25 11

### DIFF
--- a/apps/api/src/common/feature-flag.service.ts
+++ b/apps/api/src/common/feature-flag.service.ts
@@ -1,0 +1,46 @@
+import { Injectable, Logger } from '@nestjs/common';
+import * as AWS from 'aws-sdk';
+
+@Injectable()
+export class FeatureFlagService {
+  private readonly logger = new Logger(FeatureFlagService.name);
+  private readonly ssm = new AWS.SSM();
+
+  private static readonly FEATURE_FLAG_PREFIX = `/ien/${process.env.TARGET_ENV}/feature-flags`;
+
+  // we can extend here for future feature flags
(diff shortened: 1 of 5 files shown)


dbayly-freshworks (approved):
This looks good as a flag for the backend - should we also add a flag blocking access to grant reporting permissions on the front end?

jerry-ey:
@dbayly-freshworks yup, I have tested on DEV, it's also blocking access in frontend automatically. Because Frontend fetch the "/employee" endpoint to get the permission list. And we did put this feature flag under "/employee" endpoint.

This is screenshot of DEV DB and site, I have changed myself to HA user "First Nations Health Authority"
CleanShot 2024-10-10 at 14 03 03
CleanShot 2024-10-10 at 14 01 35

- You can see that HA user does not able to see the nav: Reporting
CleanShot 2024-10-10 at 14 01 50

PR #665, opened 2024-10-16
IEN-930 | IEN-938 | Automatic page refresh is clearing BCCNM upload data | Add variable pagination to the BCCNM/Inspire data review page

jerry-ey:
## Changes

- add `sessionStorage` to avoid reload clean; 
- modify page size to `5, 10 , 30, 50, 100`; (we keep `5` because of e2e testing using 5)
CleanShot 2024-10-16 at 14 38 01

- fix the previous terraform PR formatting issue to avoid the WebPR github action failure (previous PR did not have web change, hence not trigger the WebPR github action)
CleanShot 2024-10-16 at 14 11 08

## Screenshots

Have deployed to TEST and verified.
- can also verify by manually refreshing the page by clicking the refresh button on the browser.

- page size did exist before but modified to ideal number
CleanShot 2024-10-16 at 14 08 24

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -43,7 +43,7 @@ jobs:
         with:
           role-to-assume: ${{secrets.AWS_SA_ROLE_ARN}}
           aws-region: ca-central-1
-      
+
       - name: setup terraform
         uses: hashicorp/setup-terraform@v3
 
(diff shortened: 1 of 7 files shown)


dbayly-freshworks (approved):
This looks good to me!
I agree that we needed to keep 5 as a page limit. 
As a suggestion - it might make sense to pass the page sizes as a prop to the Pagination component instead of having them as constants.

jerry-ey:
Thank you, good suggestion, changed to optional props with original default PAGE_SIZES

PR #666, opened 2024-10-17
IEN-937 | Add restriction to BCCNM spreadsheet file upload

jerry-ey:
## Changes

- Add file size 6MB restriction
- Refactor the existing code: Don't need to check extension as `react-dropzone` returns empty file if the ext is not correct.

## Screenshots
Deployed to TEST and verified.
- When the ext is not correct
CleanShot 2024-10-17 at 11 03 21

- When file size > 6MB
CleanShot 2024-10-17 at 11 04 49

### DIFF
--- a/apps/web/src/components/FileUploader.tsx
+++ b/apps/web/src/components/FileUploader.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import React, { useState } from 'react';
 import { toast } from 'react-toastify';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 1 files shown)


apps/web/src/components/FileUploader.tsx
```diff
@@ -78,3 +77,27 @@ export const FileUploader = ({ extensions, handleClose, open, upload }: FileUplo
...
+    if (file.size > 6 * 1024 * 1024) {
+      throw new Error('File is too large! Please select a file smaller than 6MB');
+    }
+
+    return file;
+  } catch (err: unknown) {
+    if (err instanceof Error) {
+      toast.warning(err.message);
+    } else {
+      console.error(`An unknown error occurred: ${err}`);
```
dbayly-freshworks, line 100: Can we include a toast message here, just so the user gets a message.

dbayly-freshworks (approved):
Over all looks good. One small comment

apps/web/src/components/FileUploader.tsx
jerry-ey: good suggestion! thanks

PR #677, opened 2024-11-05
IEN-920 | End of Journey Report

jerry-ey:
- add `End of Journey` column to `milestone` extract with the following logic
```ts
/**
 * The `endOfJourneyColumn` defines a CASE statement that evaluates the journey status of an applicant.
 * It checks the `ien_applicant_status_audit` and `ien_applicant_status` tables to determine whether the applicant
 * has reached either 'End of Journey - Journey Complete' or 'End of Journey - Journey Incomplete' status.
 *
 * Logic:
 * - If there exists a record with 'End of Journey - Journey Complete' for the same `applicant_id`, the column
 *   is set to 'End of Journey - Journey Complete'.
 * - If no 'Journey Complete' record is found but a 'Journey Incomplete' record exists, the column is set to
 *   'End of Journey - Journey Incomplete'.
 * - If neither status exists for the applicant, the column value is set to NULL.
 */
```

CleanShot 2024-11-06 at 13 05 24@2x

- similarly,  add `End of Journey` column to `applicant` extract

CleanShot 2024-11-06 at 13 04 04@2x

### DIFF
--- a/apps/api/src/report/report.util.service.ts
+++ b/apps/api/src/report/report.util.service.ts
@@ -790,6 +790,41 @@ export class ReportUtilService {
       userIDString =
         'AND applicant.id IN (' + userIds.map(({ id }) => "'" + id + "'").join(',') + ')';
     }
+
+    /**
+     * The `endOfJourneyColumn` defines a CASE statement that evaluates the journey status of an applicant.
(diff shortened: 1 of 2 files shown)


dbayly-freshworks (approved):
Looks good to me

PR #684, opened 2024-11-21
IEN - 946 Sync now ignores BCCNM milestones

dbayly-freshworks:
The sync should now ignore any BCCNM specific milestones.


jerry-ey (approved):
Looks great!

PR #690, opened 2024-12-10
IEN-950 | Undo-EoJ: Implement Logic for Re-Engage milestone

jerry-ey:
NOTED: This is merging to branch `IEN-941` as it's continuing from re-engaged UI

- add 3 events: Create, Update, Delete re-engaged milestone
- These 3 events will trigger same service function which would go through complete_check and incomplete_check
- Fixed PR feedback from branch 941


apps/api/src/applicant/endofjourney.service.ts
```diff
@@ -93,8 +96,8 @@ export class EndOfJourneyService implements OnModuleInit {
     }
   }
 
-  async handleEndOfJourney<T>(
-    getter: Getter<T>,
+  async handleEndOfJourney<T, U = any>(
```
dbayly-freshworks, line 99: Any chance we can specify a type here?

apps/api/src/applicant/endofjourney.service.ts
```diff
@@ -214,20 +245,176 @@ export class EndOfJourneyService implements OnModuleInit {
...
+        'status.status AS milestone', // Include milestone (status)
+        'audit.start_date AS milestone_start_date', // Include milestone start date
+      ])
+      .leftJoin('ien_applicant_status_audit', 'audit', 'audit.applicant_id = applicant.id')
+      .leftJoin('ien_applicant_status', 'status', 'status.id = audit.status_id')
+      .where('applicant.id = :id', { id: payload.applicant.id });
+
+    const rawResults = await query.getRawMany();
+    const applicants = rawResults.reduce((result, row) => {
+      let applicant = result.find((a: any) => a.applicant_id === row.applicant_id);
```
dbayly-freshworks, line 374: Same here

dbayly-freshworks (approved):
Looks good to me, two small nits

apps/api/src/applicant/endofjourney.service.ts
jerry-ey: Thank you, fixed!

apps/api/src/applicant/endofjourney.service.ts
jerry-ey: Thank you, fixed!

PR #692, opened 2024-12-11
IEN-941 | QA fix

jerry-ey:
- Fix milestone select validation
CleanShot 2024-12-11 at 21 29 59
- Fix title when update milestone
CleanShot 2024-12-11 at 21 32 42

- Fix calendar UI/UX
CleanShot 2024-12-11 at 21 34 07

- Fix EoJ flag not being updated after CRUD re-engaged milestone

### DIFF
--- a/apps/api/src/applicant/endofjourney.service.ts
+++ b/apps/api/src/applicant/endofjourney.service.ts
@@ -251,21 +251,21 @@ export class EndOfJourneyService implements OnModuleInit {
       `Setting end of journey incomplete: ${STATUS.NOT_PROCEEDING}. Number of lists: ${list.length}`,
       'END-OF-JOURNEY',
     );
-    const results = [];
-    for (const applicant of list) {
-      // update the end_of_journey_flag and updated_date of the applicant
-      let query = this.getIncompleteQuery(manager, applicant.applicant_id);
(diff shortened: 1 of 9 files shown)


apps/api/src/applicant/endofjourney.service.ts
```diff
@@ -315,25 +315,20 @@ export class EndOfJourneyService implements OnModuleInit {
...
     this.logger.log(`Handling re-engaged event: ${event}`, 'END-OF-JOURNEY');
 
-    const connection = this.connection;
-    if (!connection) {
+    if (!manager) {
       this.logger.error('Connection failed', 'END-OF-JOURNEY');
       return;
     }
-    const queryRunner = connection.createQueryRunner();
-    await queryRunner.startTransaction();
```
dbayly-freshworks, line 330: Why are we removing the transaction here?

jerry-ey: For delete, update, create milestone in `ienapplicant.service.ts`, they all have `getManager().transaction(async manager)`, so I simply reuse this manager to avoid the stuck issue. I think 2 transactions causes the issue.

dbayly-freshworks (approved):
Looks good to me!

PR #696, opened 2024-12-24
IEN-949 | Search ATS1_ID

jerry-ey:
- Reuse the same search UI
- Add ATS1_ID to display in search result table
- Will search the ATS1_ID if the search wording is number string & length = 6.
- Fixed related unit-testing

CleanShot 2024-12-24 at 11 10 33

CleanShot 2024-12-24 at 11 10 07


dbayly-freshworks (approved):
Looks ok to me!

PR #698, opened 2025-01-08
IEN-939 | rename ncas to inspire (from user perspective)  IEN-954 | Fix broken a11y workflow

jerry-ey:
- update status name from ncas to inspire in milestone table
CleanShot 2025-01-08 at 11 03 55@2x

- update ui on Admin page
CleanShot 2025-01-08 at 11 04 25@2x

- Fix broken a11y workflow (copied same fix from EHPR)

## Note
- In an enum, we cannot use variables, so we must use plain string values. That’s why we didn’t use the variable `NCAS_NEW_NAME` from `@ien/common`.


dbayly-freshworks (approved):
Looks good!

PR #709, opened 2025-02-05
Ien 958 - Fix refresh issue 

dbayly-freshworks:
Set showHiddenApplicants to use session storage instead of use effect


jerry-ey (commented):
Looks like you miss the second requirement?

> When I am in a particular page (ex: page 20) on the Manage applicant view, the auto refresh brings be back to page 1.
> expected for the refresh to retain the page that the user is on.

jerry-ey (approved):
Approve first, so that you don't need to wait for me. Left one comment.

PR #712, opened 2025-02-18
IEN -961 Added missing message for deleted applicants

dbayly-freshworks:
Added a message for when an applicant is deleted. 

Screenshot 2025-02-18 at 12 30 29 PM

### DIFF
--- a/apps/web/src/components/applicant/ApplicantProfile.tsx
+++ b/apps/web/src/components/applicant/ApplicantProfile.tsx
@@ -94,7 +94,7 @@ export const ApplicantProfile = () => {
             </Dialog>
           </section>
         </AclMask>
-        {!applicant?.deleted_date && (
+        {!applicant?.deleted_date ? (
           <AclMask authorities={HealthAuthorities}>
             <div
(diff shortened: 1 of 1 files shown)


apps/web/src/components/applicant/ApplicantProfile.tsx
```diff
@@ -94,7 +94,7 @@ export const ApplicantProfile = () => {
             </Dialog>
           </section>
         </AclMask>
-        {!applicant?.deleted_date && (
+        {!applicant?.deleted_date ? (
```
sunghwan-park-fw, line 97: This change makes us unable to reactivate applicants. Is it ok for all users?  If it's just for HA users, we might need a condition to check the logged-in user.

dbayly-freshworks: The intent is for the applicant to be archived permanently. They are still accessible through the "show hidden applicants" toggle, and their data is still visible in the extracts.

sunghwan-park-fw (approved):

PR #736, opened 2025-09-16
Jsu 55 tf filter update

gperkinscgi:
More terraform config updates for dev deployment


npham49 (approved):
Validated locally, looks good to me!

PR #739, opened 2025-09-24
JSU-408: Fixing "Cannot delete property '0' of [object Uint8Array]" issue on AWS due to AWS considering int[] as Uint8Array(file data)

npham49:
- The error occurs because in AWS Lambda environment, certain array data gets treated as a Uint8Array rather than a regular JavaScript array. When the global ValidationPipe with [whitelist: true] tries to process this data, it attempts to delete properties from the Uint8Array, which is not allowed since Uint8Array indexed properties are immutable.
- Adding job_location to e2e test
- Updating indexOf() to includes() as per SonarQube suggestion


gperkinscgi (approved):
Looks good

PR #741, opened 2025-09-24
Bugfix for JSU-410

gperkinscgi:
Fix for JSU-410: Sorting by role


npham49 (approved):
approved and works locally!

PR #769, opened 2026-06-10
fix applicant sync date range and add diagnostics

chloe-yuu:
## Summary

Fix the applicant sync request so the ATS date range is passed correctly and add diagnostic logging to help identify where the production sync is timing out.

## Changes

- Pass `from_date` and `to_date` to the ATS applicant request flow
- Add logging for ATS applicant fetch parameters
- Add logging around milestone cleanup
- Add logging before and after marking `sync_applicants_audit` as successful

## Background

Production applicant syncs have been timing out at the Lambda 300-second limit and the "Last Sync" timestamp has remained at May 25.

CloudWatch logs show that the sync:
- Starts successfully
- Fetches applicants from ATS
- Processes applicants and milestones
- Reaches milestone cleanup
- Times out before completing and updating the sync audit record

These additional logs will help identify whether the timeout occurs during milestone cleanup or the final audit update.

## Testing

- Verified application builds successfully
- No functional changes outside applicant sync processing
- Logging changes only affect the applicant sync Lambda execution path


lucianod-bcgov (approved):
Added logs look good.

PR #770, opened 2026-06-12
fix(sync): increase sync Lambda timeout

chloe-yuu:
## Summary

Increase the `ien-prod-syncdata` Lambda timeout from 300 seconds to 900 seconds.

## Why

The sync was timing out at the previous 300-second limit before it could complete milestone cleanup and mark the audit record as successful.

A manual production run completed successfully after increasing the timeout, and the IEN Portal Last Sync value and `sync_applicants_audit` table were updated successfully.

## Notes

This makes the timeout increase permanent in Terraform so it is not reverted by a future deployment.


lucianod-bcgov (approved):
@chloe-yuu changes look good!

